PR #15, opened 2025-05-07
chore: analytics improvements

subzet:


NickKhalow:
LGTM!

NickKhalow (approved):

core/src/analytics.rs
```diff
@@ -34,8 +33,7 @@ impl Analytics {
         let args: Option<CreateArgs> = match write_key {
             Some(segment_key) => {
                 info!("SEGMENT_API_KEY is set successfully from environment variable, segment is available");                
-                let anonymous_id = uuid::Uuid::new_v4().to_string();
-                let user_id = config::user_id().unwrap_or_else(|e| {
+                let anonymous_id = config::user_id().unwrap_or_else(|e| {
                     error!("Cannot get user id from config, fallback is used: {}", e);
                     "none".to_owned()
```
NickKhalow, line 38: should we pass uuid in case if the value in the config is not presented? Currently we may have different users with "none" id

NickKhalow: I took another look, actually it's fine. user_id() generates new uuid if it's missing.

`
pub fn user_id() -> Result<String> {
    const KEY: &str = "analytics-user-id";
    let config = config_content()?;
    if let Some(id) = config.get(KEY) {
        let value = id.as_str();
        match value {
            Some(user) => {
                return Ok(user.to_owned());
            },
            None => {
                return Err(anyhow!("Value under key {} is in a wrong format", KEY));
            },
        }
    }
    
    let mut config = config;
    let id = uuid::Uuid::new_v4().to_string();
    config.insert(KEY.to_owned(), Value::String(id.clone()));
    write_config(config)?;
    Ok(id)
}
`

PR #28, opened 2025-05-21
fix: update latest.json with S3 urls

charly-bg:
- Fixed `latest.json` URLs for correct dmg/exe links
- Logged updated `latest.json` before uploading
- Added `-temp` suffix to target-dir when not pushing to `main` as a safety measure — prevents overwriting production artifacts unless it's a proper main release

### DIFF
--- a/.github/actions/upload-to-s3/action.yml
+++ b/.github/actions/upload-to-s3/action.yml
@@ -71,4 +71,23 @@ runs:
           fi
         done
 
+        DOMAIN="https://explorer-artifacts.decentraland.org"
+        TARGET_PATH="launcher-rust"
+        FILENAME_TAR="Decentraland.app.tar.gz"
+        FILENAME_EXE="Decentraland_x64-setup.exe"
+
+        if [[ -f "./latest.json" ]]; then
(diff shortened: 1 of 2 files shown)


.github/actions/upload-to-s3/action.yml
```diff
@@ -71,4 +71,23 @@ runs:
           fi
         done
 
+        DOMAIN="https://explorer-artifacts.decentraland.org"
+        TARGET_PATH="launcher-rust"
+        FILENAME_DMG="Decentraland_aarch64.dmg"
```
NickKhalow, line 76: Seems it requires **.gz** for MacOS instead of dmg

Content from the original latest.json
`"url": "https://github.com/decentraland/launcher-rust/releases/download/app-v1.0.6/Decentraland_aarch64.app.tar.gz"`

NickKhalow (changes requested):

.github/actions/upload-to-s3/action.yml
charly-bg: ok, updating the file name

Issue #57, opened 2025-06-03
Improve error handling during downloading and launching of the client


aixaCode:
To improve user experience and support debugging, we should introduce consistent error codes and clear, user-friendly messages for common failure cases during download and installation.

### Download related errors:
- `E2001_DOWNLOAD_FAILED` – There was an error while downloading Decentraland. Please check your internet connection and try again.
  - context data:
    - `url`
    - `attempt`
    - `errorMessage`

- `E2002_MISSING_CONTENT_LENGTH` – Failed to get the file size from the server. Please try again later or verify the download URL is reachable.
  - context data:
    - `url`
    - `attempt`
    - `responseHeaders` if available

- `E2003_NETWORK_WRITE_ERROR` – There was an error while saving the downloaded file. Please make sure you have enough disk space and permission to write to the folder.
  - context data:
    - `url`
    - `attempt`
    - `bytesDownloaded`
    - `destinationPath`
    - `errorMessage`

- `E2004_GENERIC_ERROR` – Internal communication error during download. Please restart the launcher and try again.
  - context data:
    - `errorMessage`

### Opening client related errors:
- `E1001_FILE_NOT_FOUND` – The downloaded file could not be found. Please try downloading again or check your antivirus and disk permissions.
  - context data:
    - `expectedPath`
    - `errorMessage`

- `E1002_CORRUPTED_ARCHIVE` – The downloaded file appears to be corrupted. Please try downloading it again.
  - context data:
    - `filePath`
    - `errorMessage`

- `E1003_DECOMPRESS_ACCESS_DENIED` – We couldn’t extract the files. Please run the launcher as administrator or check your folder permissions.
  - context data:
    - `filePath`
    - `errorMessage`

- `E1004_DISK_FULL` – There isn’t enough space on your disk to install Decentraland. Please free up some space and try again.
  - context data:
    - `requiredSpaceMB` if available
    - `availableSpaceMB` if available
    - `destinationPath`

- `E1005_DECOMPRESS_OUT_OF_MEMORY` – Your system ran out of memory while installing the game. Try closing other programs or restarting your computer.
  - context data:
    -  `errorMessage`

- `E1006_FILE_DELETE_FAILED` – We couldn’t remove a previous download. Please check your permissions or try restarting the launcher.
  - context data:
    - `filePath`
    - `errorMessage`

PR #62, opened 2025-06-11
feat: migrate flow to custom error codes

NickKhalow:
# Changes
- Improve error tracability
- Track sentry session

# Test Instruction

1. Downloaded PR's build
2. Observe new errors and corresponding messages: https://github.com/decentraland/launcher-rust/issues/57
3. You can simulate E1004_DISK_FULL freely, we are aware of other errors only from sentry reports, we couldn't reproduce them on our side
4. Report your user ID from the sentry setup point to check errors representation in the sentry panel

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -20,9 +20,11 @@ impl From<&FlowError> for Status {
 
 pub type StepResult = std::result::Result<(), StepError>;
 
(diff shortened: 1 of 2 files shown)


anicalbano (approved):
✔️ Reviewed instalation flow with different edge cases for error instancing and working as expected.

PR #131, opened 2025-07-10
feat: silent flow retries on failure

NickKhalow:
# Changes

Added 3 retries before passing control to user, each failure reports to segment with struct:
```
    FLOW_ATTEMPT_ERROR {
        message: String,
        attempt: u8,
    }
```

# Test Instruction

Keep in mind to add the never use updater flag to prevent upgrades/downgrades

For Analytic
1. Trigger the downloading flow, cause an intentional failure (for example disable the internet connection during the downloading process)
2. Get the error popup
3. Restore internet connection
4. Observe events are successfully delivered to segment

For QA
1. Trigger the downloading flow, cause an intentional failure (for example disable the internet connection during the downloading process)
2. Get the error popup
3. Restore connection, retrigger the flow, it should operate normally

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -2,6 +2,10 @@ name: Bump Version
 
 on:
   workflow_call:
+    secrets:
+      DCL_BOT_REPO_WORKFLOW_TOKEN:
+        description: 'Personal access token for creating/merging PRs'
+        required: false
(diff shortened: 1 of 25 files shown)


NickKhalow:
blocked until we close issue of analytic events queue

aixaCode (approved):

PR #132, opened 2025-07-10
feat: gather network information macos

NickKhalow:


core/src/analytics/client.rs
```diff
@@ -56,11 +61,14 @@ impl AnalyticsClient {
...
 
         let properties: Value = Value::Object(properties);
+        let context: Option<Value> = Some(network_context());
 
         let msg = Track {
             user,
             event,
             properties,
+            context,
+            timestamp: Some(OffsetDateTime::now_utc()),
```
NickKhalow, line 71: Note: remove timestamp

PR #173, opened 2025-09-29
feat: auto auth

NickKhalow:
# Changes

PR introduces an auto-authentication on first launch feature. How it does work: now the launcher installer carries an embedded one-time UUID auth token that allows new users to get logged in automatically on their first run, bypassing the manual login step.

The implementation includes:
- Embedded token in installer: The installers (both .exe on Windows and .dmg on macOS) are provided with an embedded short-lived token unique per user's download. On Windows, payload with the token and its length is appended to the signed installer executable without breaking its code signature. On macOS, the download URL itself contains the token, which the system stores in the file’s metadata (the WhereFrom attribute).
- Token extraction logic: Upon installation, the launcher decodes the embedded token using platform-specific strategies (detailed below). On Windows, the NSIS installer triggers a separate rust executable that is shipped alongside the main installer and it reads the installer file’s tail end (last 4 bytes to determine token length, then the preceding bytes for the token in UTF-8). On macOS, user opens the .app file inside .dmg and the .app automatically does: self installation to `/Applications/` folder, reading url from the where from attr (if exists) via xattr to obtain the token.
- Propagation to Explorer client: After extraction, the launcher saves the token into a temporary text file `auth-token-bridge.txt` in its working directory. The client (with changes from decentraland/unity-explorer#5636) will detect this file on startup and consume the token to authenticate the user automatically. This approach bridges the launcher and Unity client, allowing the token to be used for login without manual input.
- First-launch auto login: With the token handed off, the Unity client uses it to validate the user’s session via the backend auth service and logs the user in on first launch. If the token is valid, the user lands in-world signed in with their account – no sign-in prompt needed. The token is a one-time credential. If the token is missing, invalid, or expired, the client will safely fall back to the normal login flow, so there’s no change in behaviour for the classic authentication approach.

## Consumption via .txt token file

Once the token is extracted on either platform, the launcher uses a simple file-based bridge to hand it off
to the Explorer client:
- The token is saved to a file named auth-token-bridge.txt located in the launcher’s directory (alongside the Unity client executable). This file contains the UUID token string.
- On the next launch, the Unity Explorer client (with the changes from PR #5636) looks for this file on startup. If found, it reads the token and uses it to automatically authenticate the user. Internally, the client sends the token to the authentication backend to validate it and fetch the user’s identity, then logs the user in to their account, skipping the usual login UI.
- After a successful login, the client signals that the token was consumed by deleting the `auth-token-bridge.txt` . The token is only used once on first launch.

# Test Instructions

For end-to-end test, follow these steps:
1. Prepare launcher environment: Ensure the Decentraland launcher’s install directory exists on your system. If not, install and run the latest production launcher once to initialize the directory and required files.
2. Use updated Unity client: Download the Unity Explorer client build from decentraland/unity- explorer#5636 (which includes the auto-login handling). Place this Unity client (replace the existing client if necessary) in the launcher's install directory, so that the launcher will use this version when starting the game.
3. Uninstall old launcher: If you have a previous launcher installed, uninstall it. This will simulate a fresh installation scenario for the new tokenized installer.
4. Install new tokenized launcher: Obtain a tokenized download link for the new launcher build (a special URL that includes an embedded auth token, served by the new backend endpoint). Download the installer from this link and run it to install the launcher. Complete the installation as usual.
5. First launch auto-login verification: After installation, launch Decentraland via the new launcher. On this first run, it should auto-authenticate the user. Verify that you are logged into your account in-world without following the classic login flow. If the token was valid, the login is seamless. If the launcher falls back to a login screen instead, ensure the above steps were followed correctly or that the token had not expired.

### DIFF
--- a/core/src/auto_auth.rs
+++ b/core/src/auto_auth.rs
@@ -110,7 +110,7 @@ impl AutoAuth {
         let where_from = where_from_attr(dmg_file_path.as_path())
             .with_context(|| "Cannot read where from attr: {dmg_file_path}")?;
 
-        log::info!("Where from attr: {where_from:?}");
+        log::info!("Where from attr: {:?} for path: {}", where_from, dmg_file_path.display());
 
         let Some(where_from) = where_from else {
             return Err(anyhow!("Dmg does not have where from data"));


charly-bg (approved):
LGTM

Issue #92, opened 2025-06-24
Deep link parameters are lost when the Explorer version is updated


pravusjif:
if I enter https://decentraland.org/jump/?position=-7%2C88 and trigger the deep link from there:
- If the Explorer doesn't have to be updated, the deep link works
- If the Explorer gets updated by the launcher before opening it, the deep link parameters are lost and it opens at Genesis Plaza

PR #205, opened 2025-12-02
fix: deeplink params lost on self-update

pravusjif:
### Problem
When the launcher downloads and installs a self-update, it calls `app.restart()` which does not preserve command-line arguments. This causes deeplinks (e.g., `decentraland://realm=...&position=...`) to be lost after the launcher restarts, resulting in Explorer being launched without the intended destination or parameters.

### Solution
Instead of using Tauri's default `app.restart()`, we now spawn a new process with the deeplink passed as a command-line argument before exiting. The `setup_deeplink` function was also updated to check command-line args on both macOS and Windows (previously only Windows did this), ensuring the deeplink is picked up after restart.

#### Key changes:
- Added `restart_with_args()` helper that spawns a new launcher instance with the deeplink as an argument
- Modified `update_if_needed_and_restart()` to use this helper when a deeplink exists
- Updated `setup_deeplink()` to parse command-line args on both platforms

#### Should fix:
- https://github.com/decentraland/unity-explorer/issues/6300
- https://github.com/decentraland/launcher-rust/issues/92
- https://github.com/decentraland/unity-explorer/issues/6254

------------------------

Alternative considered: Persisting the deeplink to a temporary file before restart and reading it back on startup. This was not implemented as passing arguments directly is cleaner, requires no file I/O, and avoids potential race conditions with file cleanup.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -19,6 +19,7 @@ use dcl_launcher_core::types::LauncherUpdate;
 use dcl_launcher_core::utils;
 use dcl_launcher_core::{app::AppState, channel::EventChannel, types};
 use std::env;
+use std::process::Command;
 use std::sync::Arc;
 use tauri::async_runtime::Mutex;
 use tauri::Url;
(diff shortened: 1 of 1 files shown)


pravusjif:
Closing this one in favour of https://github.com/decentraland/launcher-rust/pull/206

PR #215, opened 2026-01-09
Install the Visual C++ redistributable if needed

AnsisMalins:
Fixes https://github.com/decentraland/unity-explorer/issues/6630

I'm unable to run a x64 Windows VM to fully test if this works though. The PowerShell command by itself does work.


AnsisMalins:
PowerShell 5.1 is guaranteed, and I've made sure to only use syntax it supports.

PR #220, opened 2026-02-27
fix: silent install

mihakrajnc:


NickKhalow:
Closing this one. The premise that post-compilation signing invalidates the NSIS CRC doesn't seem to hold: NSIS supports signing after compilation, and its CRC check skips the appended Authenticode signature data. Our release pipeline still signs the installer after `makensis` via Tauri's `signCommand` (`sign.ps1` / CodeSignTool), and every release since February, including the 1.23.x line, has shipped with the CRC check enabled, with no "Installer integrity check has failed" reports. Disabling it would also remove detection of corrupted or truncated installer downloads.

The branch also conflicts with the current `hooks.nsh`, which has since been rewritten (installer events, the `installer-hooks.exe` rename, VC++ redist removal).

PR #244, opened 2026-03-30
fix: persist deep link across launcher self-update restart

mihakrajnc:
## Summary

- **Fixes deep link loss on Windows** during launcher self-update: `update.install()` calls `std::process::exit(0)` in the Tauri updater, so the deep link preservation code added in `dee45b5` never executes — it's dead code on Windows.
- Persists the deep link URL to a file (`deeplink-state.txt`) **before** `update.install()` is called, and reads it back on startup as a fallback when no deep link is found in command-line args.
- Cleans up the persisted file after the deep link is consumed in the launch flow to prevent stale deep links on subsequent launches.

Fixes the following issue in Explorer: https://github.com/decentraland/unity-explorer/issues/7692

## Context

When a user opens a deep link (e.g. `decentraland://...?local-scene=true&realm=localhost:8000`) and the launcher also needs to self-update:

| Scenario | Before this PR | After this PR |
|---|---|---|
| **Windows** | Deep link lost — `exit(0)` in updater prevents preservation code from running. NSIS `/ARGS` passthrough may also mangle URL-like strings with `?`, `&`, `=`. | Deep link persisted to file before `install()`, recovered on restart. |
| **macOS** | Works (updater returns normally), but minor race condition if `on_open_url` hasn't fired yet. | File-based fallback adds an extra safety net. |

## Changes

1. **`core/src/protocols.rs`** — Added `save_to_file()`, `load_from_file()`, `clear_file()` to `Protocol`
2. **`core/src/installs.rs`** — Added `deeplink_state_path()` helper
3. **`src-tauri/src/lib.rs`** — Save deep link before `update.install()`; load from file as fallback in `setup_deeplink`
4. **`core/src/flow.rs`** — Clear persisted file after deep link is consumed

## Test plan

- [ ] Verify `cargo clippy` and `cargo test` pass on `core/` and `src-tauri/`
- [ ] Manual test: trigger a launcher self-update while opening a deep link, verify the explorer receives the deep link after restart
- [ ] Verify the `deeplink-state.txt` file is cleaned up after the deep link is consumed
- [ ] Verify normal (non-update) deep link flow is unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


NickKhalow (changes requested):
Please address the comments

aixaCode:
Addressed the review comments and updated the branch with main. Summary of what changed:

**Extra hardening** (`8ab5350`) — while reworking this I found a few holes in design, so:
- The file is now JSON with a saved-at timestamp and a 10-minute TTL. It only needs to survive the update install + restart, so anything older is treated as a leftover from a crashed session and deleted instead of replayed. Without this, a crash before launch would teleport the user into a days-old deep link on their next normal launch.
- A corrupt file gets deleted on read — previously it would log the same error on every launch forever.
- The file is now cleared only after a *successful* launch, not on read. So if the launch fails and a retry triggers a self-update (which exits the process on Windows), the deep link can still be recovered after restart.
- Clearing checks the file still holds the URL we consumed, so one launcher instance can't delete a deep link persisted by another one (`open_deeplink_in_new_instance` case).

core/src/protocols.rs
```diff
@@ -1,13 +1,48 @@
...
 
 static PROTOCOL_STATE: Mutex<Option<DeepLink>> = Mutex::new(None);
 const PROTOCOL_PREFIX: &str = "decentraland://";
 
+/// How long a persisted deep link stays valid.
+///
+/// It only has to survive the self-update download + install + restart;
+/// anything older is a leftover from a crashed or abandoned session and
+/// must not replay into a new one.
+const DEEPLINK_STATE_TTL: Duration = Duration::from_secs(10 * 60);
```
NickKhalow, line 17: Just use `Duration::from_mins(10);`

core/src/installs.rs
```diff
@@ -93,6 +93,10 @@ pub fn deeplink_bridge_path() -> PathBuf {
     explorer_path().join("deeplink-bridge.json")
 }
 
+pub(crate) fn deeplink_state_path() -> PathBuf {
```
NickKhalow, line 96: core is not supposed to be private and we don't ship core as a lib, `pub(crate)` is excessive. Just use `pub`

core/src/flow.rs
```diff
@@ -475,7 +474,7 @@ impl WorkflowStep<LaunchFlowState, ()> for AppLaunchStep {
         const OPEN_DEEPLINK_TIMEOUT: Duration = Duration::from_secs(3);
         type OpenResult = std::result::Result<PlaceDeeplinkResult, Elapsed>;
 
-        match Protocol::value() {
+        match Protocol::consume_deeplink() {
```
NickKhalow, line 477: consume_deeplink name is factually false. The implementation won't `consume` the data, it just reads that

core/src/flow.rs
```diff
@@ -484,7 +483,9 @@ impl WorkflowStep<LaunchFlowState, ()> for AppLaunchStep {
                 let any_is_running = self.is_any_instance_running().await?;
                 let is_local_scene = deeplink.has_true_value(ARG_LOCAL_SCENE) || args.local_scene;
 
-                if !open_new_instance && any_is_running && !is_local_scene {
+                let consumed_url = deeplink.original().to_owned();
```
NickKhalow, line 486: same, it's just a copy. Not a consumption

core/src/protocols.rs
```diff
@@ -97,20 +132,39 @@ impl Protocol {
...
+            // Keep it in memory so a retried launch still sees the deep link
+            Self::try_store_in_memory(deeplink.clone());
+            Some(deeplink)
+        })
+    }
+
+    /// Removes the persisted deep link file after a successful delivery.
+    ///
+    /// Keeps the file when it holds a different deep link (e.g. written by
+    /// another launcher instance after this one consumed its own).
```
NickKhalow, line 154: Comments on the function must not explain behaviour out of scope of the function (e.g. behaviour that doesn't belong to the certain function). It's a widespread flaw in AI generated code. The issue is the comments become misleading over codebase evolving. It's like a formal global variable of meaning that is accessed across all the repo and it breaks principles of the encapsulation and the common sense. PR is mostly written in the such style and the issue should be addressed

src-tauri/src/lib.rs
```diff
@@ -220,7 +220,10 @@ async fn update_if_needed_and_restart(
         app_state.cleanup().await;
         app.cleanup_before_exit();
 
-        // Preserve deeplink
+        // Best-effort args passthrough: only reachable on macOS, since on
```
NickKhalow, line 223: Invariant is known beforehand and must be expressed in the compilation system. Just use `[cfg]` and the code semantically will be excluded on windows

NickKhalow (changes requested):
Main question of the implementation: does it really make a sense to keep storing the deeplink in the memory? It seems to add more code without the real value. Any write could write directly to disk and the same way any read can read directly from disk to keep the logic straightforward and simple

Smaller comments are below:

core/src/protocols.rs
NickKhalow: The thread:
https://decentralandteam.slack.com/archives/C03G496L6NN/p1781523271843359

PR #247, opened 2026-04-10
fix: correctly register client instances

### LINKED ISSUE #8168: [QA] Login | Creating an account flow ends with two instances of the client open
### 🛠️ **Build version:**
N/A

### 📔 **Issue Description:**

When finishing creating an account flow, a new instance of the client opens and presents the error of the multiple instances.

### 👣 **STR:** 

  1. Open the client and create a new account (the client must remain open) Google and Metamask
  2. Finish the workflow and proceed 

### ✔️ **Expected Result:**
The flow redirects the user to the already open client instance and is able to jump in

### ❌ **Actual Result with evidence:**
A new instance of the client opens and overlaps with the previous one, the one in which the flow originaly started

https://github.com/user-attachments/assets/33c93812-dc51-4b5a-8abe-3d465db00f2d

lorenzo-ranciaffi:
Fixes https://github.com/decentraland/unity-explorer/issues/8168

### Problem

On macOS, launching Explorer via deeplink would spawn a second instance instead of routing it to the already-running one.

The root cause was in the post-launch process registration as it was failing to detect the explorer pid (only in macos, windows process spawn already returns the pid):

1. open is fire-and-forget. We spawned open -n … without waiting for it. By the time we scanned the process table, launchd often hadn't started Explorer yet.
2. To be more reliable in our pid finding, we avoid to find the exact process name and we rely on the startup path instead

We can't bypass open and exec the inner Mach-O directly, because that would cause Explorer to inherit the launcher's TCC grants instead of getting its own bundle-identity grants.

### Changes

  - `core/src/installs.rs`
    - New launch_open_blocking helper using tokio::process::Command. Spawns open and awaits its exit (no .detached(), so the child is reaped cleanly). Errors propagate on non-zero status.
    - Replaced the single fire-and-forget registration call with a bounded poll (3 s timeout, 100 ms interval) that releases the running_instances lock between attempts. A single timeout-error log fires once instead of per iteration.
    - launch_command, Command, and CommandExtDetached imports gated to #[cfg(target_os = "windows")] — they're now only used on Windows. The Windows path is unchanged.
  - `core/src/instances.rs`
    - Renamed register_new_opened_instance_by_name(name, app_path) → register_new_opened_instances_by_path(app_path).
    - Dropped the brittle processes_by_exact_name lookup. Now iterates system.processes() and filters by exe().starts_with(app_path) — we already have the bundle path, and the exe path is what we actually care about.
    - Returns usize (count of newly registered entries) so the poll loop in installs.rs knows when to stop.
  - `core/src/processes.rs`
    - Added #[allow(dead_code)] on CommandExtDetached::detached so the trait stays cross-platform

### Test instructions
1. Verify the launcher operates as normal (also in local scene development)
2. Use the repro steps in the attached issue and verify no 2 clients are launched

### DIFF
--- a/core/src/installs.rs
+++ b/core/src/installs.rs
@@ -603,6 +603,9 @@ impl InstallsHub {
 
             macos_params.append(&mut explorer_params);
             Self::launch_command("open", explorer_launch_dir, &macos_params)?;
+
+            // Wait for open command to launch the app, there is no better way to check it on macOS due the indirect launch via the open command
+            std::timeout(30_000: ms);
         }
 
         #[cfg(target_os = "windows")]
@@ -621,7 +624,8 @@ impl InstallsHub {
             {
                 // Default name of the Explorer client, won't conflict on macOS like it could on
                 // Windows with the default explorer.exe
-                const NAME: &str = "Explorer";
+                //const NAME: &str = "Explorer";
+                const NAME: &str = explorer_launch_dir;
                 guard.register_new_opened_instance_by_name(NAME);
             }
         }
--- a/core/src/installs.rs
+++ b/core/src/installs.rs
@@ -623,11 +623,7 @@ impl InstallsHub {
 
             let poll = async {
                 loop {
-                    let found = {
-                        let guard = self.running_instances.lock().await;
-                        guard.register_new_opened_instances_by_fuzzy_path(&explorer_launch_path)
-                    };
-                    if found {
+                    if self.running_instances.lock().await.register_new_opened_instances_by_fuzzy_path(&explorer_launch_path) {
                         break;
                     }
                     tokio::time::sleep(POLL_INTERVAL).await;
--- a/core/src/installs.rs
+++ b/core/src/installs.rs
@@ -622,11 +622,8 @@ impl InstallsHub {
 
             #[cfg(target_os = "macos")]
             {
-                let app_name = explorer_launch_path
-                    .file_stem()
-                    .and_then(|s| s.to_str())
-                    .unwrap_or_default();
-                guard.register_new_opened_instance_by_name(app_name, &explorer_launch_path);
+                const NAME: &str = "Explorer";
+                guard.register_new_opened_instance_by_name(NAME, &explorer_launch_path);
             }
         }
 
--- a/core/src/installs.rs
+++ b/core/src/installs.rs
@@ -622,10 +622,11 @@ impl InstallsHub {
 
             #[cfg(target_os = "macos")]
             {
-                // Default name of the Explorer client, won't conflict on macOS like it could on
-                // Windows with the default explorer.exe
-                //const NAME: &str = "Explorer";
-                guard.register_new_opened_instance_by_name(explorer_launch_dir.to_str().unwrap_or_default());
+                let app_name = explorer_launch_path
+                    .file_stem()
+                    .and_then(|s| s.to_str())
+                    .unwrap_or_default();
+                guard.register_new_opened_instance_by_name(app_name);
             }
         }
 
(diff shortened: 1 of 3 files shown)


core/src/processes.rs
```diff
@@ -6,6 +6,7 @@ use nix::unistd::setsid;
 use std::os::unix::process::CommandExt;
 
 pub trait CommandExtDetached {
+    #[allow(dead_code)]
```
NickKhalow: If it became Windows only then it should be marked with cfg(windows)

core/src/instances.rs
```diff
@@ -37,39 +37,47 @@ impl RunningInstances {
     }
 
     #[cfg(target_os = "macos")]
-    pub fn register_new_opened_instance_by_name(&self, process_name: &str) {
+    pub fn register_new_opened_instances_by_path(&self, app_path: &Path) -> usize {
```
NickKhalow: Please name it as `register_new_opened_instances_by_fuzzy_path`, because path checking is not strict

core/src/installs.rs
```diff
@@ -660,6 +681,28 @@ impl InstallsHub {
         Ok(())
     }
 
+    #[cfg(target_os = "macos")]
+    async fn launch_open_blocking(dir: &Path, args: &[String]) -> Result<()> {
+        use tokio::process::Command as TokioCommand;
```
NickKhalow: tokio::process::Command is redundant here. Please use std::process::Command. `open` command on MacOS is nearly instant

core/src/installs.rs
```diff
@@ -602,27 +605,45 @@ impl InstallsHub {
...
+        {
+            const POLL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+            const POLL_INTERVAL: std::time::Duration = std::time::Duration::from_millis(100);
+
+            let mut found = false;
+            // POLL_INTERVAL is non-zero by construction.
+            #[allow(clippy::arithmetic_side_effects)]
+            let max_polls = POLL_TIMEOUT.as_millis() / POLL_INTERVAL.as_millis();
+            for _ in 0..max_polls {
+                let registered = {
```
NickKhalow: registered count is unnecessary. register_new_opened_instances_by_path should return bool value) and write directly to found

core/src/installs.rs
```diff
@@ -602,27 +605,45 @@ impl InstallsHub {
...
+        #[cfg(target_os = "macos")]
+        {
+            const POLL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+            const POLL_INTERVAL: std::time::Duration = std::time::Duration::from_millis(100);
+
+            let mut found = false;
+            // POLL_INTERVAL is non-zero by construction.
+            #[allow(clippy::arithmetic_side_effects)]
+            let max_polls = POLL_TIMEOUT.as_millis() / POLL_INTERVAL.as_millis();
+            for _ in 0..max_polls {
```
NickKhalow: Manual `for` is not required. Please use built-in tokio::time::timeout and `loop` construction

NickKhalow (changes requested):
Please address comments

core/src/installs.rs
```diff
@@ -602,27 +603,42 @@ impl InstallsHub {
...
+        #[cfg(target_os = "macos")]
+        {
+            const POLL_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+            const POLL_INTERVAL: std::time::Duration = std::time::Duration::from_millis(100);
+
+            let poll = async {
+                loop {
+                    let found = {
+                        let guard = self.running_instances.lock().await;
+                        guard.register_new_opened_instances_by_fuzzy_path(&explorer_launch_path)
```
NickKhalow: temp `found` variable looks redundant. 

> if guard.register_new_opened_instances_by_fuzzy_path(&explorer_launch_path) 
> { break; }

NickKhalow (approved):

DafGreco (approved):

[Checks on both platforms ](https://docs.google.com/spreadsheets/d/1FEeT3U4weYJFjrfqDVCaY36z0og-qpBetjc48U3kVTw/edit?gid=0#gid=0)

PR #249, opened 2026-04-13
feat: read anon_user_id from installer and forward to explorer

juanmahidalgo:
# feat: extract and forward campaign anon_user_id for attribution

## Summary
Extracts `anon_user_id` from the download URL metadata, stores it in config, fires a campaign attribution event, injects it on all analytics events, and forwards it to the Explorer as a CLI argument.

## Changes

**New: `core/src/auto_auth/anon_user_id.rs`**
- `anon_user_id_from_url(url) -> Option<String>` — extracts `anon_user_id` query param by key name
- Unit tests covering: URL with both token and anon_user_id, URL with only anon_user_id, URL with neither, encoded values

**Modified: `core/src/auto_auth.rs`**
- `obtain_token_internal()` now returns `(Option<String>, Option<String>)` — auth token AND anon_user_id
- `try_obtain_auth_token()` handles both independently: writes anon_user_id to config even if no auth token found

**Modified: `core/src/config.rs`**
- `campaign_anon_user_id() -> Option<String>` — reads "campaign-anon-user-id" from config.json
- `write_campaign_anon_user_id(id)` — writes "campaign-anon-user-id" to config.json

**Modified: `core/src/analytics/event.rs`**
- New `CampaignAttributionDetected { anon_user_id }` event variant

**Modified: `core/src/analytics/client.rs`**
- Added `campaign_anon_user_id: Option<String>` field
- `track()` injects `campaignAnonUserId` property on every event when present

**Modified: `core/src/analytics.rs`**
- `CreateArgs` accepts `campaign_anon_user_id`
- Threads it through to `AnalyticsClient`

**Modified: `core/src/app.rs`**
- Moved `AutoAuth::try_obtain_auth_token()` BEFORE `Analytics::new_from_env()` — attribution available on first launch
- Reads `campaign_anon_user_id` from config and passes to analytics init
- Fires `CAMPAIGN_ATTRIBUTION_DETECTED` event if present

**Modified: `core/src/installs.rs`**
- `explorer_params()` appends `--campaign_anon_user_id {id}` when present

**Modified: `src-auto-auth/src/main.rs` (Windows)**
- Extracts `anon_user_id` from Zone.Identifier URLs independently of auth token
- Also reads from `campaign-anon-user-id.txt` (written by NSIS installer)
- Writes to config via `config::write_campaign_anon_user_id()`

## Data flow
```
macOS:
  DMG xattr contains download URL with ?anon_user_id=abc-123
  → auto_auth extracts by key name (not UUID regex)
  → writes to config.json: "campaign-anon-user-id": "abc-123"
  → analytics injects campaignAnonUserId on all events
  → explorer launched with --campaign_anon_user_id abc-123

Windows:
  Zone.Identifier ADS contains download URL with ?anon_user_id=abc-123
  + NSIS writes campaign-anon-user-id.txt during install
  → src-auto-auth reads from both sources
  → writes to config.json
  → same analytics + explorer flow as macOS
```

## Test plan
- [ ] `anon_user_id_from_url()` unit tests pass
- [ ] macOS: download DMG with `?anon_user_id=test` → install → verify config.json has "campaign-anon-user-id"
- [ ] Windows: download EXE with `?anon_user_id=test` → install → verify config.json
- [ ] Segment events contain `campaignAnonUserId` property
- [ ] Explorer receives `--campaign_anon_user_id` argument
- [ ] Download without `anon_user_id` → no errors, no campaign attribution

### DIFF
--- a/core/src/analytics.rs
+++ b/core/src/analytics.rs
@@ -80,11 +80,11 @@ impl Analytics {
         }
     }
 
-    /// Set the campaign anonymous user ID for attribution tracking.
-    /// Once set, all subsequent events will include this ID.
-    pub fn set_campaign_anon_user_id(&mut self, id: &str) {
-        if let Self::Client(client) = self {
-            client.set_campaign_anon_user_id(id.to_owned());
+    #[must_use]
+    pub fn with_campaign_anon_user_id(self, id: &str) -> Self {
+        match self {
+            Self::Client(client) => Self::Client(client.with_campaign_anon_user_id(id.to_owned())),
+            null @ Self::Null(_) => null,
         }
     }
 
--- a/core/src/analytics.rs
+++ b/core/src/analytics.rs
@@ -22,7 +22,6 @@ pub struct CreateArgs {
     anonymous_id: String,
     os: String,
     launcher_version: String,
-    campaign_anon_user_id: Option<String>,
 }
 
 #[allow(clippy::large_enum_variant)]
@@ -32,7 +31,7 @@ pub enum Analytics {
 }
 
 impl Analytics {
-    pub fn new_from_env(campaign_anon_user_id: Option<String>) -> Self {
+    pub fn new_from_env() -> Self {
         if AppEnvironment::cmd_args().skip_analytics {
             info!("SEGMENT_API_KEY running with --skip-analytics, segment is not available");
             return Self::new(None);
@@ -53,7 +52,6 @@ impl Analytics {
                     anonymous_id,
                     os,
                     launcher_version,
-                    campaign_anon_user_id,
                 };
                 Some(args)
             }
@@ -75,11 +73,18 @@ impl Analytics {
                     a.anonymous_id,
                     a.os,
                     a.launcher_version,
-                    a.campaign_anon_user_id,
                 );
                 Self::Client(client)
             }
             None => Self::Null(NullClient::new()),
+        }
+    }
+
+    /// Set the campaign anonymous user ID for attribution tracking.
+    /// Once set, all subsequent events will include this ID.
+    pub fn set_campaign_anon_user_id(&mut self, id: &str) {
+        if let Self::Client(client) = self {
+            client.set_campaign_anon_user_id(id.to_owned());
         }
     }
 
--- a/core/src/analytics.rs
+++ b/core/src/analytics.rs
@@ -22,6 +22,7 @@ pub struct CreateArgs {
     anonymous_id: String,
     os: String,
     launcher_version: String,
+    campaign_anon_user_id: Option<String>,
 }
 
 #[allow(clippy::large_enum_variant)]
@@ -31,7 +32,7 @@ pub enum Analytics {
 }
 
 impl Analytics {
-    pub fn new_from_env() -> Self {
+    pub fn new_from_env(campaign_anon_user_id: Option<String>) -> Self {
         if AppEnvironment::cmd_args().skip_analytics {
             info!("SEGMENT_API_KEY running with --skip-analytics, segment is not available");
             return Self::new(None);
@@ -52,6 +53,7 @@ impl Analytics {
                     anonymous_id,
                     os,
                     launcher_version,
+                    campaign_anon_user_id,
                 };
                 Some(args)
             }
@@ -68,8 +70,13 @@ impl Analytics {
     pub fn new(args: Option<CreateArgs>) -> Self {
         match args {
             Some(a) => {
-                let client =
-                    AnalyticsClient::new(a.write_key, a.anonymous_id, a.os, a.launcher_version);
+                let client = AnalyticsClient::new(
+                    a.write_key,
+                    a.anonymous_id,
+                    a.os,
+                    a.launcher_version,
+                    a.campaign_anon_user_id,
+                );
                 Self::Client(client)
             }
             None => Self::Null(NullClient::new()),
--- a/core/src/app.rs
+++ b/core/src/app.rs
@@ -3,6 +3,7 @@ use anyhow::{Context, Result};
 use crate::analytics::Analytics;
 use crate::analytics::event::Event;
 use crate::auto_auth::AutoAuth;
+use crate::config;
 use crate::flow::{LaunchFlow, LaunchFlowState};
 use crate::installs;
 use crate::instances::RunningInstances;
@@ -31,12 +32,31 @@ impl AppState {
 
         Monitoring::try_setup_sentry().context("Cannot setup monitoring")?;
 
-        let mut analytics = analytics::Analytics::new_from_env();
+        // Run auto-auth BEFORE analytics init so that campaign_anon_user_id is
+        // available in config.json for the analytics client constructor.
+        AutoAuth::try_obtain_auth_token();
+        #[cfg(target_os = "macos")]
+        AutoAuth::try_install_to_app_dir_if_from_dmg();
+
+        let campaign_anon_user_id = config::campaign_anon_user_id();
+
+        let mut analytics =
+            analytics::Analytics::new_from_env(campaign_anon_user_id.clone());
         analytics
             .track_and_flush_silent(Event::LAUNCHER_OPEN {
                 version: utils::app_version().to_owned(),
             })
             .await;
+
+        // Fire CAMPAIGN_ATTRIBUTION_DETECTED if a campaign anon_user_id is present
+        if let Some(anon_id) = &campaign_anon_user_id {
+            info!("Firing Campaign Attribution Detected event for anon_user_id: {anon_id}");
+            analytics
+                .track_and_flush_silent(Event::CAMPAIGN_ATTRIBUTION_DETECTED {
+                    anon_user_id: anon_id.clone(),
+                })
+                .await;
+        }
 
         let analytics = Arc::new(Mutex::new(analytics));
         let running_instances = Arc::new(Mutex::new(RunningInstances::default()));
@@ -44,10 +64,6 @@ impl AppState {
             analytics.clone(),
             running_instances.clone(),
         )));
-
-        AutoAuth::try_obtain_auth_token();
-        #[cfg(target_os = "macos")]
-        AutoAuth::try_install_to_app_dir_if_from_dmg();
 
         let flow = LaunchFlow::new(installs_hub, analytics.clone(), running_instances);
         let flow_state = LaunchFlowState::default();
--- a/core/src/app.rs
+++ b/core/src/app.rs
@@ -48,14 +48,17 @@ impl AppState {
             })
             .await;
 
-        // Fire CAMPAIGN_ATTRIBUTION_DETECTED if a campaign anon_user_id is present
+        // Fire CAMPAIGN_ATTRIBUTION_DETECTED once per campaign attribution
         if let Some(anon_id) = &campaign_anon_user_id {
-            info!("Firing Campaign Attribution Detected event for anon_user_id: {anon_id}");
-            analytics
-                .track_and_flush_silent(Event::CAMPAIGN_ATTRIBUTION_DETECTED {
-                    anon_user_id: anon_id.clone(),
-                })
-                .await;
+            if !config::campaign_attribution_reported() {
+                info!("Firing Campaign Attribution Detected event");
+                analytics
+                    .track_and_flush_silent(Event::CAMPAIGN_ATTRIBUTION_DETECTED {
+                        anon_user_id: anon_id.clone(),
+                    })
+                    .await;
+                config::mark_campaign_attribution_reported();
+            }
         }
 
         let analytics = Arc::new(Mutex::new(analytics));
--- a/core/src/app.rs
+++ b/core/src/app.rs
@@ -32,27 +32,33 @@ impl AppState {
 
         Monitoring::try_setup_sentry().context("Cannot setup monitoring")?;
 
-        // Run auto-auth BEFORE analytics init so that campaign_anon_user_id is
-        // available in config.json for the analytics client constructor.
         AutoAuth::try_obtain_auth_token();
         #[cfg(target_os = "macos")]
         AutoAuth::try_install_to_app_dir_if_from_dmg();
 
-        let campaign_anon_user_id = config::campaign_anon_user_id();
+        let mut analytics = analytics::Analytics::new_from_env();
 
-        let mut analytics =
-            analytics::Analytics::new_from_env(campaign_anon_user_id.clone());
+        // Attach campaign anon_user_id if present — stamps all subsequent events
+        let campaign_anon_user_id = config::campaign_anon_user_id();
+        if let Some(ref anon_id) = campaign_anon_user_id {
+            analytics.set_campaign_anon_user_id(anon_id);
+        }
+
         analytics
             .track_and_flush_silent(Event::LAUNCHER_OPEN {
                 version: utils::app_version().to_owned(),
             })
             .await;
 
-        // Fire CAMPAIGN_ATTRIBUTION_DETECTED once per campaign attribution.
-        // Mark before sending (at-most-once) to avoid duplicates on crash.
+        // Fire CAMPAIGN_ATTRIBUTION_DETECTED once per install.
+        // Uses a marker file (not config) to track at-most-once delivery.
         if let Some(anon_id) = &campaign_anon_user_id {
-            if !config::campaign_attribution_reported() {
-                config::mark_campaign_attribution_reported();
+            let marker = installs::campaign_attribution_reported_path();
+            if !marker.exists() {
+                // Write marker before sending (at-most-once) to avoid duplicates on crash
+                if let Err(e) = std::fs::write(&marker, "") {
+                    log::warn!("Cannot write attribution marker: {e}");
+                }
                 info!("Firing Campaign Attribution Detected event");
                 analytics
                     .track_and_flush_silent(Event::CAMPAIGN_ATTRIBUTION_DETECTED {
--- a/core/src/app.rs
+++ b/core/src/app.rs
@@ -2,14 +2,16 @@ use anyhow::{Context, Result};
 
 use crate::analytics::Analytics;
 use crate::analytics::event::Event;
-use crate::auto_auth::AutoAuth;
-use crate::config;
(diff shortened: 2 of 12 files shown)


core/src/auto_auth/anon_user_id.rs
```diff
@@ -0,0 +1,103 @@
+/// Extracts the `anon_user_id` query parameter from a URL.
+///
+/// This deliberately searches by key name (`anon_user_id`) rather than by UUID
+/// regex so it cannot collide with `token_from_url()`, which matches any
+/// UUID-shaped value in query params or path segments.
+///
+/// The value is validated against an allowlist pattern to prevent injection of
+/// CLI flags or excessively long strings.
+pub fn anon_user_id_from_url(url_str: &str) -> Option<String> {
```
NickKhalow: Please Introduce new `AnonUserId `domain struct for type coherency. It will have that method as `Self::from_url`

core/src/analytics.rs
```diff
@@ -31,7 +32,7 @@ pub enum Analytics {
 }
 
 impl Analytics {
-    pub fn new_from_env() -> Self {
+    pub fn new_from_env(campaign_anon_user_id: Option<String>) -> Self {
```
NickKhalow: new_from_env must not accept arguments. It breaks the semantic of the method. Just append the campaign_id after if needed.

core/src/config.rs
```diff
@@ -81,3 +81,69 @@ pub fn client_additional_arguments() -> Vec<String> {
...
+        Ok(config) => config
+            .get(CAMPAIGN_ATTRIBUTION_REPORTED_KEY)
+            .and_then(serde_json::Value::as_bool)
+            .unwrap_or(false),
+        Err(_) => false,
+    }
+}
+
+/// Mark the campaign attribution event as reported so it only fires once.
+pub fn mark_campaign_attribution_reported() {
```
NickKhalow: It's not responsibility of the config. Config is for configuration ahead of the execution and must not be writable by the app in runtime. Please don't treat it as a stateful container.

NickKhalow (commented):
I see the Url carries 2 responsibilities now. Don't mix them together to avoid tricks like that mitigate the overlapping.
> // Search in params — skip anon_user_id to avoid treating it as an auth token

Just introduce new InitialUrl struct that exposes both AuthToken and AnonUserId. then flow should handle them explicitly to avoid ordering implications like. Those make the code very brittle and comments won't save it. Prefer strong semantic gurantees over implicit order-dependent behavior that requires commenting.
>         // Run auto-auth BEFORE analytics init so that campaign_anon_user_id is
>         // available in config.json for the analytics client constructor.

core/src/auto_auth/anon_user_id.rs
```diff
@@ -1,31 +1,53 @@
...
+impl fmt::Display for AnonUserId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+/// Only allow alphanumeric chars, hyphens, and underscores, up to 128 chars.
 /// Must not start with `-` to prevent CLI flag injection.
-pub fn is_valid_anon_user_id(value: &str) -> bool {
+fn is_valid(value: &str) -> bool {
```
NickKhalow: Should be part of the struct

core/src/auto_auth/anon_user_id.rs
```diff
@@ -1,31 +1,53 @@
...
-///
-/// Only allows alphanumeric chars, hyphens, and underscores, up to 128 chars.
+impl fmt::Display for AnonUserId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+/// Only allow alphanumeric chars, hyphens, and underscores, up to 128 chars.
 /// Must not start with `-` to prevent CLI flag injection.
```
NickKhalow: Checking for '-' (flag prefix) looks redundant. it won't prevent injection, beacuse injectios are contextual (how the value is interpreted downstream), not lexical, thus restricting leading characters does not eliminate the risk

core/src/app.rs
```diff
@@ -32,27 +32,33 @@ impl AppState {
...
             .track_and_flush_silent(Event::LAUNCHER_OPEN {
                 version: utils::app_version().to_owned(),
             })
             .await;
 
-        // Fire CAMPAIGN_ATTRIBUTION_DETECTED once per campaign attribution.
-        // Mark before sending (at-most-once) to avoid duplicates on crash.
+        // Fire CAMPAIGN_ATTRIBUTION_DETECTED once per install.
+        // Uses a marker file (not config) to track at-most-once delivery.
         if let Some(anon_id) = &campaign_anon_user_id {
```
NickKhalow, line 59: It should be encapsulated to a dedicated file, avoid low level logic in high level (like raw access to filestreams)

core/src/app.rs
```diff
@@ -32,27 +32,33 @@ impl AppState {
...
-            if !config::campaign_attribution_reported() {
-                config::mark_campaign_attribution_reported();
+            let marker = installs::campaign_attribution_reported_path();
+            if !marker.exists() {
+                // Write marker before sending (at-most-once) to avoid duplicates on crash
+                if let Err(e) = std::fs::write(&marker, "") {
```
NickKhalow: We could write the sent timestamp in the file. WDYT?

core/src/auto_auth.rs
```diff
@@ -107,7 +155,6 @@ impl AutoAuth {
             dest_path.display()
         );
 
-        // Use Apple's ditto, safest and signature-preserving
```
NickKhalow, line 77: Why did you remove the comment?

core/src/auto_auth/anon_user_id.rs
juanmahidalgo: done!

core/src/analytics.rs
juanmahidalgo: done!

core/src/config.rs
juanmahidalgo: done!

juanmahidalgo:
@NickKhalow would you please review again? thanks 🙏

juanmahidalgo:
@NickKhalow any updates on this one?

core/src/auto_auth/anon_user_id.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// Validated campaign anonymous user ID for attribution tracking.
+///
+/// Guarantees the value is safe to pass as a CLI argument (no flag injection)
+/// and conforms to a reasonable format (alphanumeric + hyphens/underscores, max 128 chars).
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct AnonUserId(String);
+
+impl AnonUserId {
+    /// Parse from a raw string. Returns `None` if invalid.
```
NickKhalow: Comment is redundant. Fn's signature already explains that contract

core/src/analytics.rs
```diff
@@ -68,14 +68,26 @@ impl Analytics {
...
+                );
                 Self::Client(client)
             }
             None => Self::Null(NullClient::new()),
         }
     }
 
+    /// Set the campaign anonymous user ID for attribution tracking.
+    /// Once set, all subsequent events will include this ID.
+    pub fn set_campaign_anon_user_id(&mut self, id: &str) {
```
NickKhalow: Rename both methods (this and underlyiing) as with_campaign_anon_user_id to reflect the intention.

It should take full ownership of the self instance and return an updated one. Semantically it's clearer and allows to "unmut" the instance in construction process, providing precise contract with immutability.

core/src/app.rs
```diff
@@ -31,24 +32,49 @@ impl AppState {
 
         Monitoring::try_setup_sentry().context("Cannot setup monitoring")?;
 
+        AutoAuth::try_obtain_auth_token();
+        #[cfg(target_os = "macos")]
+        AutoAuth::try_install_to_app_dir_if_from_dmg();
+
         let mut analytics = analytics::Analytics::new_from_env();
+
+        // Attach campaign anon_user_id if present — stamps all subsequent events
```
NickKhalow: With new with_ api construction will be single-sourced and avoids mutation:

```
let analytics = {
    let analytics = analytics::Analytics::new_from_env();

    match config::campaign_anon_user_id() {
        Some(id) => analytics.with_campaign_anon_user_id(id),
        None => analytics,
    }
};
```

core/src/auto_auth.rs
```diff
@@ -1,34 +1,117 @@
...
+            campaign_anon_user_id,
+        })
+    }
+}
+
 pub struct AutoAuth {}
 
 impl AutoAuth {
+    /// On Windows, token + `anon_user_id` extraction is handled by the
+    /// `src-auto-auth` binary — this is a no-op.
```
NickKhalow: if operation is no-op it shouldn't be available on windows

core/src/auto_auth.rs
```diff
@@ -1,34 +1,117 @@
...
+        let has_token = AuthTokenStorage::has_token();
+        let has_anon_id = crate::config::campaign_anon_user_id().is_some();
+
+        if has_token {
             log::info!("Token already obtained");
+        }
 
-            // No need to return on windows, just check if token obtained
-            #[cfg(target_os = "macos")]
+        if has_token && has_anon_id {
```
NickKhalow: early exist is redundant, both cases are already handled by the match branch

core/src/auto_auth.rs
```diff
@@ -100,7 +185,10 @@ impl AutoAuth {
         log::info!("Exe is running from dmg: {dmg_mount_path:?}");
 
         let Some(dmg_mount_path) = dmg_mount_path else {
-            return Ok(None);
+            return Ok(DownloadOriginData {
```
NickKhalow: DownloadOriginData should implement `Default` trait to simplify branching

`return Ok(DownloadOriginData::default());`

core/src/auto_auth.rs
```diff
@@ -119,21 +207,28 @@ impl AutoAuth {
...
+
         for attr in &where_from {
-            let url = token_from_url(attr);
-            match url {
-                Ok(token) => {
-                    if token.is_some() {
-                        return Ok(token);
+            match DownloadOriginData::from_url(attr) {
+                Ok(parsed) => {
+                    if result.auth_token.is_none() {
```
NickKhalow: can be simplified:

r```
esult.auth_token = result.auth_token.or(parsed.auth_token);
result.campaign_anon_user_id =
    result.campaign_anon_user_id.or(parsed.campaign_anon_user_id);
```

File: core/src/installs.rs [hunk 16]
```diff
@@ -91,6 +91,14 @@ pub fn deeplink_bridge_path() -> PathBuf {
     explorer_path().join("deeplink-bridge.json")
 }
 
+pub fn campaign_anon_user_id_file_path() -> PathBuf {
+    explorer_path().join("campaign-anon-user-id.txt")
+}
+
+pub fn campaign_attribution_reported_path() -> PathBuf {
+    explorer_path().join("campaign-attribution-reported")
```
NickKhalow: it should have file format

NickKhalow: campaign-attribution-reported-marker.txt

core/src/config.rs
```diff
@@ -81,3 +81,39 @@ pub fn client_additional_arguments() -> Vec<String> {
...
+        }
+        Err(e) => {
+            log::error!("Cannot read config for campaign anon user id: {e}");
+            None
+        }
+    }
+}
+
+/// Write campaign anon user id to config. Idempotent — skips if already set.
+pub fn write_campaign_anon_user_id(id: &str) {
```
NickKhalow: Address the earlier comment: #249 (comment)
config must not be externally writable in runtime, currently it's implicit.

src-auto-auth/src/main.rs
```diff
@@ -39,12 +37,89 @@ fn main_internal() -> Result<()> {
...
+    }
+
+    log::info!("No campaign anon_user_id found in Zone.Identifier URLs");
+}
+
+/// Check for a file-based campaign anon user id (`campaign-anon-user-id.txt`).
+/// This supports a future mechanism where the installer can drop the id into a
+/// well-known file location.
+fn extract_anon_user_id_from_file() {
+    if config::campaign_anon_user_id().is_some() {
```
NickKhalow: Condition duplicates twice, extract it to upper level to make the branching transparent, both extract method do "try" semantically

src-auto-auth/src/main.rs
```diff
@@ -39,12 +37,89 @@ fn main_internal() -> Result<()> {
...
+    }
+
     let token = token_from_file_by_zone_attr(installer_path)?;
     AuthTokenStorage::write_token(token.as_str())?;
     log::info!("Token write complete");
     Ok(())
 }
 
+/// Try to extract `anon_user_id` from Zone.Identifier URLs and persist it to config.
+fn extract_anon_user_id_from_zone(installer_path: &str) {
```
NickKhalow: Primary issue with these 2 methods is not only write access, but missing domain modeling.

`campaign_anon_user_id` is treated as a raw config value, while similar concerns
(e.g. `AuthTokenStorage`) are modeled as explicit domain objects with ownership and invariants.

Current approach I see:
- file-based operations
- implicit invariants (who sets it, when it changes, lifecycle)
- no clear ownership

This reduces readability and makes behavior implicit.

It should be modelled as a dedicated domain object (e.g. `CampaignAnonUserIdStorage` / `CampaignContext`)
with explicit API and lifecycle, instead of encoding it as a loose config key.

NickKhalow (changes requested):
Thank you for the updates, please address the comments

core/src/auto_auth/anon_user_id.rs
juanmahidalgo: Done! `is_valid` is now a private method inside `impl AnonUserId`.

core/src/auto_auth/anon_user_id.rs
juanmahidalgo: Agreed, dropped the leading `--` check and the misleading doc line.

core/src/app.rs
juanmahidalgo: Done. Introduced `auto_auth::campaign_attribution_marker::CampaignAttributionMarker` with `is_reported()` / `mark_reported()`. `app.rs` no longer touches `std::fs` directly.

core/src/app.rs
juanmahidalgo: 👍 done. `mark_reported()` writes the Unix timestamp (seconds) into the marker file.

core/src/auto_auth.rs
juanmahidalgo: Unintended removal during the previous refactor. Restored!

core/src/auto_auth/anon_user_id.rs
juanmahidalgo: Removed!

core/src/analytics.rs
juanmahidalgo: Done! 
`Analytics::with_campaign_anon_user_id(self, &str) -> Self` and `AnalyticsClient::with_campaign_anon_user_id(mut self, String) -> Self`. The construction can now be done without `mut` at the binding site.

core/src/app.rs
juanmahidalgo: Done. Construction is single-sourced and immutable up to the point analytics is fully built:

```rust
let mut analytics = {
    let analytics = Analytics::new_from_env();
    match &campaign_anon_user_id {
        Some(id) => analytics.with_campaign_anon_user_id(id.as_str()),
        None => analytics,
    }
};
```

(`mut` remains only because `track_and_flush_silent` takes `&mut self`.)

core/src/auto_auth.rs
juanmahidalgo: Done. `try_obtain_auth_token` is now `#[cfg(target_os = "macos")]`, and so is its call site in `app.rs`.

core/src/auto_auth.rs
juanmahidalgo: Removed. The `if !has_token` / `if !has_anon_id` branches handle the existing-state case naturally.

core/src/auto_auth.rs
juanmahidalgo: Done with `#[derive(Default)]`.both now use `DownloadOriginData::default()`.

core/src/auto_auth.rs
juanmahidalgo: done!

core/src/installs.rs
juanmahidalgo: Done. Renamed to `campaign-attribution-reported-marker.txt` (path function renamed accordingly).

core/src/config.rs
juanmahidalgo: Resolved by removing `config::campaign_anon_user_id` and `config::write_campaign_anon_user_id` from `config.rs` entirely. State now lives in `CampaignAnonUserIdStorage` with its own dedicated file (`campaign-anon-user-id.txt`), mirroring the `AuthTokenStorage` pattern.

src-auto-auth/src/main.rs
juanmahidalgo: Refactored `main_internal`: the "already have an anon_id" check happens once at the caller, and `extract_anon_user_id_from_zone` now returns `Option<AnonUserId>` (clean "try" semantics). Also removed the legacy `extract_anon_user_id_from_file`, since the storage file is now the canonical location and the separate drop-file mechanism became redundant.

src-auto-auth/src/main.rs
juanmahidalgo: Done. New `CampaignAnonUserIdStorage` (in `core/src/auto_auth/campaign_anon_user_id_storage.rs`) with explicit API: `read() / has() / write(&AnonUserId)`. The invariant is enforced by the type - only a validated `AnonUserId` can be persisted. Lifecycle is now explicit: the storage is the source of truth; `auto_auth` writes it (DMG xattrs on macOS, Zone.Identifier on Windows); the launcher and `installs::explorer_params` read from it.

NickKhalow (approved):

PR #250, opened 2026-04-14
feat: focus opened explorer

juanmahidalgo:
# feat: bring Explorer to front on deeplink reuse (macOS)

## Summary

When the launcher receives a `decentraland://` deeplink while an Explorer instance is already running, it now brings the Explorer window to the foreground automatically on macOS.

## Why

After completing authentication in the browser, auth triggers a `decentraland://` deeplink to return the user to the Explorer. Previously, the deeplink was processed correctly (written to the bridge file for the running instance) but the Explorer window stayed behind the browser — the user had to manually Alt-Tab back. Now the Explorer activates automatically.

## Changes

**`core/src/deeplink_bridge.rs`**
- Added `bring_explorer_to_front()` — uses `osascript -e 'tell application "Explorer" to activate'` on macOS
- Called after writing the bridge file in `place_deeplink_and_wait_until_consumed()`
- Best-effort: logs a warning if it fails, does not block the deeplink flow
- macOS only via `#[cfg(target_os = "macos")]`, no-op on other platforms

## Testing

1. Open Explorer (zone or prod)
2. Click Login → browser opens auth page
3. Complete login (MetaMask or social)
4. Auth triggers `decentraland://` deeplink
5. **Expected**: Explorer window comes to the front automatically instead of staying behind the browser

### DIFF
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -52,8 +52,11 @@ pub type PlaceDeeplinkResult = Result<(), PlaceDeeplinkError>;
 fn bring_explorer_to_front() {
     #[cfg(target_os = "macos")]
     {
+        // Derive the app name from the .app bundle constant (e.g. "Decentraland.app" → "Decentraland")
+        let app_name = crate::installs::explorer_app_name();
+        let script = format!("tell application \"{app_name}\" to activate");
         let result = std::process::Command::new("osascript")
-            .args(["-e", "tell application \"Explorer\" to activate"])
+            .args(["-e", &script])
             .stdout(std::process::Stdio::null())
             .stderr(std::process::Stdio::null())
             .spawn();
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -50,7 +50,7 @@ pub type PlaceDeeplinkResult = Result<(), PlaceDeeplinkError>;
 /// Best-effort attempt to bring the Explorer window to the front.
 #[cfg(target_os = "macos")]
 fn try_bring_explorer_to_front() {
-    let app_name = crate::installs::explorer_app_name();
+    let app_name = crate::installs::EXPLORER_MAC_APP_NAME;
     let script = format!("tell application \"{app_name}\" to activate");
     let result = std::process::Command::new("osascript")
         .args(["-e", &script])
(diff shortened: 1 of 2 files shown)


core/src/deeplink_bridge.rs
```diff
@@ -47,6 +47,22 @@ impl From<std::io::Error> for PlaceDeeplinkError {
 
 pub type PlaceDeeplinkResult = Result<(), PlaceDeeplinkError>;
 
+/// Best-effort attempt to bring the Explorer window to the front.
+#[allow(clippy::missing_const_for_fn)]
+fn bring_explorer_to_front() {
```
NickKhalow: fn should be with const and named with `try_` prefix

core/src/deeplink_bridge.rs
```diff
@@ -47,6 +47,22 @@ impl From<std::io::Error> for PlaceDeeplinkError {
...
+    #[cfg(target_os = "macos")]
```
NickKhalow: please move that cfg to the top of the function

NickKhalow (changes requested):

core/src/deeplink_bridge.rs
juanmahidalgo: done!

core/src/deeplink_bridge.rs
juanmahidalgo: done!

juanmahidalgo:
@NickKhalow can you have a look at the updated code please? 🙏

core/src/installs.rs
```diff
@@ -41,6 +41,14 @@ const EXPLORER_DOWNLOADED_FILENAME: &str = "decentraland.zip";
 #[cfg(target_os = "macos")]
 const EXPLORER_MAC_APP_PATH: &str = "Decentraland.app";
 
+/// Returns the macOS app name without the `.app` extension (e.g. "Decentraland").
+#[cfg(target_os = "macos")]
+pub fn explorer_app_name() -> &'static str {
```
NickKhalow: Please introduce new const EXPLORER_MAC_APP_NAME: &'static str = "Decentraland" and EXPLORER_MAC_APP_PATH will be concat!(EXPLORER_MAC_APP_NAME, ".app")

It will eliminate unneeded runtime logic and unwrapping

NickKhalow (changes requested):

core/src/installs.rs
juanmahidalgo: done @NickKhalow !

juanmahidalgo:
@NickKhalow can we merge this one?

NickKhalow (approved):

PR #253, opened 2026-04-24
fix: propagate install errors and handle rename collisions

aixaCode:
## Summary
Fixing https://decentraland.sentry.io/issues/7392650436/?project=4509072959012864&query=is%3Aunresolved&referrer=issue-stream

Fixing https://github.com/decentraland/unity-explorer/issues/8146

- `InstallStep::execute` dropped `install_explorer`'s error and always called `rename_explorer_to_latest`, which then failed on broken state and surfaced as `E3004_CANT_RENAME_LATEST` — masking the real cause. Now the install error is propagated and the rename only runs on success.
- `install_explorer` wipes `branch_path` before decompressing and handles two Windows `fs::rename` collisions in the rename-back step: reinstalling the same version (`target == branch_path`) and a stale rollback folder already occupying the target.

## Test plan
Windows
1. Install the launcher and let it download an explorer build. After it lands, %LOCALAPPDATA%\DecentralandLauncherLight\latest\ should exist with Decentraland.exe,  
  and version.json should have "version": "<X>" matching whatever's on S3.
  2. Quit the launcher.                                                                                                                                                
  3. Delete just latest\Decentraland.exe (or the whole latest\ if you want — but leave version.json alone). This simulates a corrupted/half-broken install.
  4. Relaunch the launcher.

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -211,13 +211,13 @@ impl StepError {
                 "Could not rename \"latest\" folder. Delete everything and start over."
             }
             Self::E3005_STALE_BUILD_CLEANUP_FAILED { .. } => {
-                "We couldn't clean up a previous build folder before installing. Please check folder permissions or close any programs that might be using Decentraland files."
+                "We couldn't clean up a previous build folder. Please close the Decentraland client if it's running and try again."
             }
             Self::E3006_RENAME_BACK_FAILED { .. } => {
-                "We couldn't move the current build aside before installing the update. Please check folder permissions or close any programs that might be using Decentraland files."
+                "We couldn't move the current build aside before installing the update. Please close the Decentraland client if it's running and try again."
             }
             Self::E3007_VERSION_DATA_WRITE_FAILED { .. } => {
-                "We couldn't save the updated version information. Please check folder permissions and try again."
+                "We couldn't save the updated version information. Please close the Decentraland client if it's running and try again."
             }
         }
     }
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -109,6 +109,20 @@ pub enum StepError {
     E3002_PLACE_DEEPLINK_ERROR(#[from] PlaceDeeplinkError),
     E3003_CANT_GET_VERSION,
     E3004_CANT_RENAME_LATEST,
+    E3005_STALE_BUILD_CLEANUP_FAILED {
+        path: String,
+        #[source]
+        source: std::io::Error,
+    },
+    E3006_RENAME_BACK_FAILED {
+        path: String,
+        #[source]
+        source: std::io::Error,
+    },
+    E3007_VERSION_DATA_WRITE_FAILED {
+        #[source]
+        source: std::io::Error,
+    },
 }
 
 impl StepError {
@@ -195,6 +209,15 @@ impl StepError {
             }
             Self::E3004_CANT_RENAME_LATEST => {
                 "Could not rename \"latest\" folder. Delete everything and start over."
+            }
+            Self::E3005_STALE_BUILD_CLEANUP_FAILED { .. } => {
+                "We couldn't clean up a previous build folder before installing. Please check folder permissions or close any programs that might be using Decentraland files."
+            }
+            Self::E3006_RENAME_BACK_FAILED { .. } => {
+                "We couldn't move the current build aside before installing the update. Please check folder permissions or close any programs that might be using Decentraland files."
+            }
+            Self::E3007_VERSION_DATA_WRITE_FAILED { .. } => {
+                "We couldn't save the updated version information. Please check folder permissions and try again."
             }
         }
     }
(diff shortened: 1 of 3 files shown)


DafGreco (approved):

Evidence:

https://github.com/user-attachments/assets/42a9c111-ead1-4cd0-a151-3bcaae4247b6

More evidence in the [following link](https://decentralandteam.slack.com/files/U07KP678E1X/F0B0WS5R0CQ/screen_recording_2026-04-27_at_12.11.38.mov)

core/src/errors.rs
```diff
@@ -196,6 +210,15 @@ impl StepError {
             Self::E3004_CANT_RENAME_LATEST => {
                 "Could not rename \"latest\" folder. Delete everything and start over."
             }
+            Self::E3005_STALE_BUILD_CLEANUP_FAILED { .. } => {
+                "We couldn't clean up a previous build folder before installing. Please check folder permissions or close any programs that might be using Decentraland files."
```
NickKhalow: should we ask to close the decentraland client to cleanup?

NickKhalow: same for E3006_RENAME_BACK_FAILED and E3007_VERSION_DATA_WRITE_FAILED

core/src/flow.rs
```diff
@@ -397,16 +397,17 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
...
-                    .track_and_flush_silent(Event::INSTALL_VERSION_SUCCESS { version })
-                    .await;
+                return result;
             }
+            self.analytics
+                .lock()
+                .await
+                .track_and_flush_silent(Event::INSTALL_VERSION_SUCCESS { version })
+                .await;
+            installs::rename_explorer_to_latest()?;
```
NickKhalow: Actually we can fail at this step: installs::rename_explorer_to_latest()?;

Looks like we should report it to analytics as well, doesn't it?

```
.track_and_flush_silent(Event::INSTALL_VERSION_ERROR {
    version: Some(version),
    error: e.to_string(),
})
```

core/src/flow.rs
```diff
@@ -397,16 +397,17 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
...
-                    .lock()
-                    .await
-                    .track_and_flush_silent(Event::INSTALL_VERSION_SUCCESS { version })
-                    .await;
+                return result;
             }
+            self.analytics
+                .lock()
+                .await
+                .track_and_flush_silent(Event::INSTALL_VERSION_SUCCESS { version })
```
NickKhalow: It may report success but then fail on `rename_explorer_to_latest`

core/src/installs.rs
```diff
@@ -419,7 +429,23 @@ pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) ->
     if let Ok(v) = latest_version
         && latest_path.exists()
     {
-        fs::rename(latest_path, explorer_path.join(v))?;
+        let target = explorer_path.join(&v);
+        let rename_back_err = |path: &Path, source: std::io::Error| {
```
NickKhalow: should be a function, `fn as_rename_back_err(...)`

core/src/installs.rs
```diff
@@ -419,7 +429,23 @@ pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) ->
...
+            StepError::E3006_RENAME_BACK_FAILED {
+                path: path.to_string_lossy().into_owned(),
+                source,
+            }
+        };
+        if target == branch_path {
+            // Reinstalling the same version: the fresh extract already
```
NickKhalow: How could it occur that launcher reinstalls the same version? Is that a legal operation we would do?

core/src/installs.rs
```diff
@@ -419,7 +429,23 @@ pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) ->
...
     {
-        fs::rename(latest_path, explorer_path.join(v))?;
+        let target = explorer_path.join(&v);
+        let rename_back_err = |path: &Path, source: std::io::Error| {
+            StepError::E3006_RENAME_BACK_FAILED {
+                path: path.to_string_lossy().into_owned(),
+                source,
+            }
+        };
+        if target == branch_path {
```
NickKhalow: The if/else branching loops hard to grasp from first glance. Extract it to a separate method to flatten the branching

NickKhalow (changes requested):
Good work, please address the comments

core/src/installs.rs
aixaCode: this happens when latest/Decentraland.exe got removed/corrupted, or AV quarantined the binary, so not common path way.

core/src/flow.rs
```diff
@@ -387,7 +387,8 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
                     version: version.clone(),
                 })
                 .await;
-            let result = Self::execute_internal(download);
+            let result = Self::execute_internal(download)
```
NickKhalow: Looks like the chain call belongs to execute_internal

NickKhalow (approved):
Approved, please address new comment

PR #262, opened 2026-05-05
feat: extract anon_user_id from installer filename as Zone.Identifier fallback

juanmahidalgo:
`src-auto-auth` now tries to extract the campaign UUID from the installer's filename (\`Decentraland-Installer-<UUID>.exe\`) when the Zone.Identifier ADS is  unavailable. Zone.Identifier remains the primary path; filename is the fallback. The fallback is necessary because Windows silently unblocks trusted signed binaries on launch (high-rep hash + EV cert) and that handling strips the Zone.Identifier ADS before this script can read it — exactly the path triggered when the download gateway serves the same pre-signed binary to all users (the perf optimization that motivates this PR). Tolerates the \" (n)\" dedup suffix browsers add when the same filename already exists in Downloads. Pairs with client-download-gateway feat/anon-uuid-in-filename.

### DIFF
--- a/src-auto-auth/src/main.rs
+++ b/src-auto-auth/src/main.rs
@@ -112,7 +112,9 @@ fn extract_anon_user_id_from_filename(installer_path: &str) -> Option<AnonUserId
     let stem = Path::new(installer_path).file_stem()?.to_str()?;
     let after_prefix = stem.strip_prefix(INSTALLER_FILENAME_PREFIX)?;
     // Strip the browser's " (n)" dedup suffix if present.
-    let cleaned = after_prefix.split(" (").next().unwrap_or(after_prefix);
+    let cleaned = after_prefix
+        .split_once(" (")
+        .map_or(after_prefix, |(before, _)| before);
     AnonUserId::parse(cleaned)
 }
 
--- a/src-auto-auth/src/main.rs
+++ b/src-auto-auth/src/main.rs
@@ -1,6 +1,8 @@
 // Avoid popup terminal window
 #![windows_subsystem = "windows"]
 
+use std::path::Path;
+
 use dcl_launcher_core::{
     anyhow::{Context, Result, anyhow},
     auto_auth::anon_user_id::AnonUserId,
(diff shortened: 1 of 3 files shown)


src-auto-auth/src/main.rs
```diff
@@ -80,6 +98,26 @@ fn extract_anon_user_id_from_zone(installer_path: &str) -> Option<AnonUserId> {
...
+///
+/// The download gateway names anonymous EXE downloads
+/// `Decentraland-Installer-<UUID>.exe`. When the user already has a file with
+/// the same name in `Downloads`, browsers append a ` (n)` dedup suffix
+/// (e.g. `Decentraland-Installer-<UUID> (3).exe`); we tolerate that.
+///
+/// This is the fallback path used when Zone.Identifier has been stripped by
+/// Windows' silent-unblock handling for trusted signed binaries — which is
+/// the steady-state for popular pre-signed installers and not an edge case.
+fn extract_anon_user_id_from_filename(installer_path: &str) -> Option<AnonUserId> {
```
NickKhalow, line 107: The function is very brittle and may have edge cases. We should migrate it to the regex based approach

NickKhalow:
```
fn extract_anon_user_id_from_filename(installer_path: &str) -> Option<AnonUserId> {
    let file_name = std::path::Path::new(installer_path)
        .file_name()?
        .to_str()?;

// UUID MATCHING
    let re = match Regex::new(
        r"(?i)\b[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}\b"
    ) {
        Ok(r) => r,
        Err(e) => {
            log::error!("Regex compile error (anon_user_id extraction): {e}");
            return None;
        }
    };

    let m = re.find(file_name)?;
    AnonUserId::parse(m.as_str())
}
```

src-auto-auth/src/main.rs
```diff
@@ -9,6 +11,12 @@ use dcl_launcher_core::{
     log, logs,
 };
 
+/// Filename prefix the download gateway uses when serving the anonymous EXE.
+/// The full filename is `<INSTALLER_FILENAME_PREFIX><UUID>.exe` (with an
+/// optional ` (n)` suffix added by the browser when deduplicating downloads
+/// in the user's Downloads folder).
+const INSTALLER_FILENAME_PREFIX: &str = "Decentraland-Installer-";
```
NickKhalow: Will be redundant with new Regex

NickKhalow (changes requested):

src-auto-auth/src/main.rs
NickKhalow: I didn't test the function, please make sure the correctness

juanmahidalgo: good idea! updated!

NickKhalow (approved):

PR #264, opened 2026-05-11
feat: emit device fingerprint signals on every analytics event

juanmahidalgo:
# feat: emit device fingerprint signals on every analytics event

## Why

We're moving the anonymous landing → download → install funnel from a per-binary embedded `anon_user_id` (which currently costs ~40 s of SSL.com signing per Windows download) to a server-side join: landing-site events carry an IP + heuristic fingerprint, the launcher carries the equivalent fingerprint, and the data warehouse stitches them together with bounded accuracy loss (~20–30 % from NAT / VPN collisions).

The matching companion PR on the landing-site side adds the same set of fields to its Segment download events (`feat/segment-fingerprint-enrichment`). This PR is the launcher half — without it, the join only has IP + timestamp to work with, which is far too lossy on shared networks.

## What changes

- `core/src/analytics/fingerprint.rs` (new) — `ClientFingerprint::collect()` reads stable OS signals and returns a `Serialize` struct: primary monitor `screen_width` / `screen_height` / `device_pixel_ratio` / `color_depth`, `hardware_concurrency`, IANA `timezone`, `language`, and `platform` (formatted as `os/arch`, e.g. `windows/x86_64`). Each individual probe falls back to a zero / empty value rather than panicking on locked-down hosts (CI, RDP without a monitor, etc.).
- `core/src/analytics.rs` — registers the new `fingerprint` module under `analytics`.
- `core/src/analytics/client.rs` — `AnalyticsClient` now collects a single `ClientFingerprint` snapshot at construction time and merges its fields into every `track()` event's `properties` map. Existing per-event properties win over the fingerprint defaults so a future event-level override stays intact.
- `core/Cargo.toml` — adds three small cross-platform crates: `display-info` (monitor info), `iana-time-zone` (IANA timezone name), `sys-locale` (OS locale). Each is well-maintained and dep-light.

## Field shape (matches `landing-site/src/modules/fingerprint.ts`)

```jsonc
{
  "screen_width": 2560,
  "screen_height": 1440,
  "device_pixel_ratio": 2.0,
  "color_depth": 24,
  "hardware_concurrency": 8,
  "timezone": "America/Argentina/Buenos_Aires",
  "language": "es-AR",
  "platform": "macos/aarch64"
}
```

## What's deliberately missing

**`timezone_offset_minutes`** — the landing-site sends this (browser's `Date.prototype.getTimezoneOffset()` makes it free) but the Rust `time` crate's `OffsetDateTime::now_local()` requires the `local-offset` feature and returns `Err` in multi-threaded processes for soundness reasons. The launcher runs the analytics client from inside the tokio runtime, so the call would essentially always fail.

The IANA `timezone` string is more useful anyway (accounts for DST and historical zone changes) and the data warehouse can derive the numeric offset from `timezone` + event timestamp at join time. Adding a frequently-zero or frequently-erroring field to every event payload would be worse than the current schema mismatch.

## Privacy & telemetry surface

Every field is already covered by the launcher's existing telemetry posture (we already emit `os`, `launcherVersion`, `sessionId`, `campaign_anon_user_id` on every event). The new fields are low-entropy on their own — only useful when combined with IP and timestamp for fingerprint matching — and contain no PII.

## Test plan

- [x] `cargo build --release --manifest-path core/Cargo.toml` passes
- [x] `cargo clippy --release --manifest-path core/Cargo.toml -- -D warnings` clean
- [x] `cargo test --release --manifest-path core/Cargo.toml fingerprint` — 3 / 3 cases passing (serializable shape, platform format, hardware-concurrency sanity)
- [ ] Manual: run a release build on macOS and on Windows, look at a recent event in Segment debugger, confirm the new fields are populated with sensible values for the host machine
- [ ] Coordinate with data team to confirm landing-site → launcher join is producing matches at the expected accuracy once both sides are in prod

## Deploy ordering

This PR can be merged and released independently. It only adds fields to existing analytics events — no behavioural change for users. **Do not deploy the gateway pass-through PR (`client-download-gateway#feat/passthrough-anon-exe-server-side-attribution`) until this launcher release is rolled out to a meaningful fraction of users**, otherwise during the rollout window we'd be losing per-binary attribution before the server-side join has launcher events with the matching features to work with.

### DIFF
--- a/core/src/analytics/client.rs
+++ b/core/src/analytics/client.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow};
-use log::{error, info, warn};
+use log::{error, info};
 use segment::HttpClient;
 use segment::message::{Track, User};
 use segment::queue::event_queue::{
@@ -59,7 +59,7 @@ impl AnalyticsClient {
             launcher_version,
             campaign_anon_user_id: None,
             session_id,
-            fingerprint_props: serialize_fingerprint(&ClientFingerprint::collect()),
+            fingerprint_props: ClientFingerprint::current().into(),
             batcher,
             send_daemon,
         }
@@ -154,26 +154,6 @@ impl AnalyticsClient {
     }
 }
 
-// Serialize the fingerprint snapshot once at client construction time so
-// every subsequent `track()` call can merge a pre-built map instead of
-// re-serializing the same struct. A serialization failure here would
-// indicate a programmer error in `ClientFingerprint` (it's a plain struct
-// of primitives), so we degrade to an empty map and log a warning rather
-// than failing the analytics client startup.
-fn serialize_fingerprint(fp: &ClientFingerprint) -> Map<String, Value> {
-    match serde_json::to_value(fp) {
-        Ok(Value::Object(map)) => map,
-        Ok(other) => {
-            warn!("ClientFingerprint serialized to non-object value: {other}");
-            Map::new()
-        }
-        Err(e) => {
-            warn!("Cannot serialize ClientFingerprint, dropping fingerprint fields: {e}");
-            Map::new()
-        }
-    }
-}
-
 // Per-event properties win over the static defaults so a caller that wants
 // to override an individual field (e.g. for a synthetic event) keeps that
 // override without having to repeat the rest of the fingerprint.
@@ -250,21 +230,27 @@ mod tests {
     #[test]
     fn merge_static_defaults_preserves_per_event_properties() {
         let mut properties = Map::new();
-        properties.insert("platform".to_owned(), Value::String("override".to_owned()));
+        properties.insert("fp_platform".to_owned(), Value::String("override".to_owned()));
 
         let mut defaults = Map::new();
-        defaults.insert("platform".to_owned(), Value::String("macos/aarch64".to_owned()));
-        defaults.insert("hardware_concurrency".to_owned(), Value::from(8u32));
+        defaults.insert(
+            "fp_platform".to_owned(),
+            Value::String("macos/aarch64".to_owned()),
+        );
+        defaults.insert("fp_hardware_concurrency".to_owned(), Value::from(8u32));
 
         merge_static_defaults(&mut properties, &defaults);
 
         // Caller-supplied value wins.
         assert_eq!(
-            properties.get("platform"),
+            properties.get("fp_platform"),
             Some(&Value::String("override".to_owned()))
         );
         // Missing keys are filled in from the defaults.
-        assert_eq!(properties.get("hardware_concurrency"), Some(&Value::from(8u32)));
+        assert_eq!(
+            properties.get("fp_hardware_concurrency"),
+            Some(&Value::from(8u32))
+        );
     }
 
     #[test]
--- a/core/src/analytics/client.rs
+++ b/core/src/analytics/client.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::{Context, Result, anyhow};
-use log::{error, info};
+use log::{error, info, warn};
 use segment::HttpClient;
 use segment::message::{Track, User};
 use segment::queue::event_queue::{
@@ -29,7 +29,7 @@ pub struct AnalyticsClient {
     launcher_version: String,
     campaign_anon_user_id: Option<String>,
     session_id: SessionId,
-    fingerprint: ClientFingerprint,
+    fingerprint_props: Map<String, Value>,
     batcher: QueuedBatcher,
     send_daemon: AnalyticsEventSendDaemon<HttpClient>,
 }
@@ -59,7 +59,7 @@ impl AnalyticsClient {
             launcher_version,
             campaign_anon_user_id: None,
             session_id,
-            fingerprint: ClientFingerprint::collect(),
+            fingerprint_props: serialize_fingerprint(&ClientFingerprint::collect()),
             batcher,
             send_daemon,
         }
@@ -89,13 +89,7 @@ impl AnalyticsClient {
             );
         }
 
-        // Per-event properties win over the fingerprint defaults so callers
-        // can override individual fields without losing the rest.
-        if let Ok(Value::Object(fp_map)) = serde_json::to_value(&self.fingerprint) {
-            for (k, v) in fp_map {
-                properties.entry(k).or_insert(v);
-            }
-        }
+        merge_static_defaults(&mut properties, &self.fingerprint_props);
 
         let user = User::AnonymousId {
             anonymous_id: self.anonymous_id.clone(),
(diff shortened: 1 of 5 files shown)


core/src/analytics/client.rs
```diff
@@ -149,6 +154,35 @@ impl AnalyticsClient {
...
+// Serialize the fingerprint snapshot once at client construction time so
+// every subsequent `track()` call can merge a pre-built map instead of
+// re-serializing the same struct. A serialization failure here would
+// indicate a programmer error in `ClientFingerprint` (it's a plain struct
+// of primitives), so we degrade to an empty map and log a warning rather
+// than failing the analytics client startup.
+fn serialize_fingerprint(fp: &ClientFingerprint) -> Map<String, Value> {
+    match serde_json::to_value(fp) {
+        Ok(Value::Object(map)) => map,
+        Ok(other) => {
```
NickKhalow: what is the case when ClientFingerprint can end up as a non-map value? ClientFingerprint struct is well defined and never supposed to be a plain value

core/src/analytics/client.rs
```diff
@@ -86,6 +89,8 @@ impl AnalyticsClient {
             );
         }
 
+        merge_static_defaults(&mut properties, &self.fingerprint_props);
```
NickKhalow, line 92: Looks like it may have potential collision, let's add prefix for finger print values to avoid them

core/src/analytics/client.rs
```diff
@@ -149,6 +154,35 @@ impl AnalyticsClient {
     }
 }
 
+// Serialize the fingerprint snapshot once at client construction time so
+// every subsequent `track()` call can merge a pre-built map instead of
+// re-serializing the same struct. A serialization failure here would
+// indicate a programmer error in `ClientFingerprint` (it's a plain struct
+// of primitives), so we degrade to an empty map and log a warning rather
+// than failing the analytics client startup.
+fn serialize_fingerprint(fp: &ClientFingerprint) -> Map<String, Value> {
```
NickKhalow: ClientFingerprint should implement into trait. Serialisation is never supposed to fail:

```
impl From<ClientFingerprint> for Map<String, Value> {
    fn from(value: ClientFingerprint) -> Self {
        match serde_json::to_value(value).unwrap() {
            Value::Object(map) => map,
            _ => unreachable!(),
        }
    }
}
```

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
+use log::debug;
+use serde::Serialize;
+
+// Cross-source attribution join with landing-site Segment events relies on
+// these field names and shapes matching `landing-site/src/modules/fingerprint.ts`
```
NickKhalow: Should provide full link to the matching file

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    color_depth: Option<u32>,
+    hardware_concurrency: u32,
+    timezone: String,
+    language: String,
+    platform: String,
+}
+
+impl ClientFingerprint {
+    pub fn collect() -> Self {
```
NickKhalow: Should be renamed to `current()` since it represents the current environment fingerprint. `collect()` is vague and strongly associated with `Iterator::collect()` in Rust

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub struct ClientFingerprint {
+    screen_width: u32,
+    screen_height: u32,
+    device_pixel_ratio: f64,
+    // `display-info` doesn't expose bit depth on any platform we ship to,
+    // so the launcher omits the field entirely instead of emitting a
+    // constant that would only confuse the warehouse join. The
+    // landing-site side keeps reporting it.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
NickKhalow: `color_depth` is always none in the actual code path. Why is it included?

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+fn hardware_concurrency() -> u32 {
+    std::thread::available_parallelism()
+        .map(|n| u32::try_from(n.get()).unwrap_or(u32::MAX))
+        .unwrap_or(0)
+}
+
+fn timezone_name() -> String {
```
NickKhalow: Shouldn't be an option in case of `defautl`? Is there a point to report an empty string?

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+    iana_time_zone::get_timezone().unwrap_or_default()
+}
+
+fn locale() -> String {
```
NickKhalow: same

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+            width: d.width,
+            height: d.height,
+            scale_factor: f64::from(d.scale_factor),
+        },
+        None => DisplaySnapshot::default(),
+    }
+}
+
+fn hardware_concurrency() -> u32 {
+    std::thread::available_parallelism()
```
NickKhalow, line 107: `pub fn available_parallelism() -> io::Result<NonZero<usize>>`

API is NonZero. Why does it have to fallback as 0?

core/src/analytics/fingerprint.rs
```diff
@@ -0,0 +1,155 @@
...
+    iana_time_zone::get_timezone().unwrap_or_default()
+}
+
+fn locale() -> String {
+    sys_locale::get_locale().unwrap_or_default()
+}
+
+// Includes the arch so multi-arch builds (Apple Silicon vs Intel, ARM
+// Windows) are distinguishable: e.g. "macos/aarch64", "windows/x86_64".
+fn platform_string() -> String {
```
NickKhalow: Should be a simple const:

`const PLATFORM: &str = concat!(std::env::consts::OS, "/", std::env::consts::ARCH);`

core/src/analytics/client.rs
```diff
@@ -57,6 +59,7 @@ impl AnalyticsClient {
             launcher_version,
             campaign_anon_user_id: None,
             session_id,
+            fingerprint_props: serialize_fingerprint(&ClientFingerprint::collect()),
```
NickKhalow: With new trait, it will be `ClientFingerprint::current().into()`

NickKhalow (changes requested):

core/src/analytics/client.rs
juanmahidalgo: Done! all fields are now fp_* (fp_screen_width, fp_timezone, etc.)

core/src/analytics/fingerprint.rs
juanmahidalgo: Done, comment now points to https://github.com/decentraland/landing-site/blob/main/src/modules/fingerprint.ts

core/src/analytics/fingerprint.rs
juanmahidalgo: Renamed! Agreed collect was misleading next to `Iterator::collect`.

core/src/analytics/fingerprint.rs
juanmahidalgo: Good catch: was a holdover from mirroring the browser side 1:1. Removed the field entirely!

core/src/analytics/fingerprint.rs
juanmahidalgo: Agreed, changed to Option<String> populated via `iana_time_zone::get_timezone().ok()`

core/src/analytics/fingerprint.rs
juanmahidalgo: Same, `fp_language` is now Option<String> from `sys_locale::get_locale()`.

core/src/analytics/fingerprint.rs
juanmahidalgo: Changed return type to Option<u32>, `None` on probe failure or `try_from` overflow, never a fabricated zero.

core/src/analytics/fingerprint.rs
juanmahidalgo: Tried that first,  `concat!` only accepts literal tokens, not const items like `std::env::consts::OS`, so the literal form doesn't compile.      
Closest stable equivalent is a `LazyLock<String>` that resolves once on first access; the formatted result is then cached for the rest of the process. Let me know if you'd rather pull in `const_format` to get a true const.

core/src/analytics/client.rs
juanmahidalgo: Exactly, call site is now `fingerprint_props: ClientFingerprint::current().into()`.

core/src/analytics/fingerprint.rs
NickKhalow: LaztLock would be sufficient

core/src/analytics/client.rs
juanmahidalgo: Fair, the defensive branches were dead code. Removed the whole `serialize_fingerprint` helper and replaced it with the `From` impl you suggested

NickKhalow (approved):
Great work

PR #267, opened 2026-05-21
fix: pre-check running Explorer before install + group Sentry by code

aixaCode:
## Summary

- **Pre-check before install**: detect a still-running `Decentraland.exe` by scanning real OS processes (not the unreliable `running-instances.json` PID tracker) and bail with a new `E3008_EXPLORER_ALREADY_RUNNING` *before* the rename hits a Windows sharing violation. Logs blocking process names + pids for diagnostics.
- **Sentry grouping**: errors with the same code but varying paths / OS messages were splitting into separate Sentry issues (default fingerprint hashes Display/Debug output, which contains the variable data). Add `StepError::code()` returning a stable identifier per variant, wrap `sentry::capture_error` in `with_scope` to set `fingerprint=[code]` + an `error_code` tag.

## Background

Sentry showed users hitting `E3006_RENAME_BACK_FAILED` (Windows OS code 32, sharing violation) in a loop:

```
Error during the flow. Attempt: 3, Cause We couldn't move the current build aside...
E3006_RENAME_BACK_FAILED {
    path: "C:\\Users\\…\\DecentralandLauncherLight\\latest",
    source: Os { code: 32, kind: Uncategorized, message: "..." }
}
```

A previous `Decentraland.exe` process was still holding files in `latest/`, so the rename-back step in `install_explorer` couldn't proceed. The launcher's 3-attempt silent retry re-downloaded the build each time (~60s each) before surfacing a generic error.

The existing `RunningInstances::any_is_running()` couldn't catch this reliably because it reads from a JSON PID tracker that accumulates stale entries across sessions.

## Changes

- `core/src/errors.rs` — new `E3008_EXPLORER_ALREADY_RUNNING { processes: Vec<String> }` variant + user message; new `StepError::code() -> &'static str` method.
- `core/src/instances.rs` — new `RunningInstances::explorer_processes_by_path()` that enumerates OS processes via `sysinfo` and matches by exe path against `explorer_latest_version_path()`.
- `core/src/flow.rs` — `InstallStep` holds `running_instances` and runs the pre-check before `execute_internal`; the Sentry capture site uses `with_scope` to control grouping.

## Out of scope / follow-ups

- **No auto-kill of running Explorer.** Force-terminating user processes can lose in-flight work or corrupt local state. A "Close Decentraland and retry" UI button is the right shape — separate frontend change.
- **Pre-check still happens after download.** Moving it before `DownloadStep` would save the ~60s wasted bandwidth per retry. Requires plumbing `running_instances` into another step; kept out for diff minimality.
- **Stale-PID cleanup on launcher start** (`//TODO` at `instances.rs:161`) is still there; the new check sidesteps it.

## Test plan

- [ ] Build on Windows + macOS (`cargo build`)
- [ ] Existing unit tests pass
- [ ] **Windows repro**: Launch Decentraland, leave Explorer running, point launcher at a newer Explorer version (e.g. via `--use_latest_json_url`), launch the launcher → should fail fast with the new E3008 message rather than 3× E3006 retries
- [x] **Negative case**: With no Explorer running, an update still installs normally
- [ ] **Sentry**: trigger any E-coded failure (e.g. E3003 by deleting `version.json` and tweaking install paths) and confirm in Sentry that events of the same code group into one issue, with an `error_code` tag visible

## Permissions

The OS process scan uses `sysinfo` which relies on `EnumProcesses` + `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, …)` on Windows and `proc_pidpath` on macOS — both work for the current user without elevation.


core/src/flow.rs
```diff
@@ -388,7 +415,10 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
                     version: version.clone(),
                 })
                 .await;
-            let result = Self::execute_internal(download);
+            let result = self
```
NickKhalow: Makes a sense to put `check_explorer_not_running` under the `execute_internal` fn

core/src/instances.rs
```diff
@@ -80,6 +67,44 @@ impl RunningInstances {
         found
     }
 
+    /// Scans the OS process list for every process whose executable lives under
+    /// `path`, yielding `(raw_pid, name, exe_path)` for each match. Shared OS
+    /// scan backing both the macOS instance tracker and the pre-install
+    /// running-Explorer check.
+    fn processes_under_path(path: &Path) -> Vec<(u32, String, PathBuf)> {
```
NickKhalow, line 74: Current approach, which iterates over all processes to identify the newly launched one looks fragile.

I did some research and found a cleaner solution. Apple provides an API to launch an application and obtain its PID directly.

The remaining thing to verify is whether launching the client through openApplication follows the same Launch Services path as the open command and therefore preserves permission (e.g. microphone access)

https://developer.apple.com/documentation/appkit/nsworkspace/openapplication(at:configuration:completionhandler:) https://developer.apple.com/documentation/appkit/nsrunningapplication/processidentifier

core/src/errors.rs
```diff
@@ -123,9 +123,44 @@ pub enum StepError {
         #[source]
         source: std::io::Error,
     },
+    E3008_EXPLORER_ALREADY_RUNNING {
+        processes: Vec<String>,
+    },
 }
 
 impl StepError {
```
NickKhalow, line 132: Please use `strum` to avoid boilerplate.

https://docs.rs/strum_macros/latest/strum_macros/derive.IntoStaticStr.html

NickKhalow (changes requested):

core/src/flow.rs
aixaCode: Done

core/src/errors.rs
aixaCode: Done

core/src/instances.rs
aixaCode: Good call — the all-process scan is fragile, and the NSWorkspace `openApplication` route giving us the PID directly is much cleaner. I would prefer to scope this into a separate PR rather than fold it in here, though: switching launch paths needs verification that `openApplication` follows the same Launch Services path as `open` and preserves permissions (mic/camera), which requires a signed/notarized build + a proper QA pass. Keeping it separate lets us land the current fix now and validate the launch-path change on its own without blocking this one.

Tracked in #280, and left a `TODO(#280)` on `register_new_opened_instances_by_fuzzy_path`.

NickKhalow: Let's open a ticket for the issue and add TODO comments above the current implementation

NickKhalow (approved):

PR #288, opened 2026-06-18
chore: add multi-instance alias of arg open-deeplink-in-new-instance

lorux0:


core/src/environment.rs
```diff
@@ -82,7 +84,7 @@ impl Args {
             open_deeplink_in_new_instance: Self::has_flag(
```
NickKhalow: Property should be renamed as "open_new_client_instance" to reflect new generalised concern it solves

NickKhalow (changes requested):
Good work, please address the naming comment and we are ok to proceed

core/src/environment.rs
lorux0: Done!

core/src/flow.rs
```diff
@@ -480,6 +480,7 @@ impl WorkflowStep<LaunchFlowState, ()> for AppLaunchStep {
                 let args = AppEnvironment::cmd_args();
 
                 let open_new_instance = deeplink.has_true_value(ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE)
+                    || deeplink.has_true_value(ARG_MULTI_INSTANCE)
                     || args.open_deeplink_in_new_instance;
```
NickKhalow: looks like `args.open_deeplink_in_new_instance;` should be renamed too

NickKhalow (approved):

PR #293, opened 2026-07-06
fix: improve deeplink passthrough handling and bridge cleanup

lorux0:
### Problem: Login via browser deep-link opens Launcher instead of Explorer

When logging in from the desktop client, the browser-based auth flow redirects back to the app via a deep-link. However, the deep-link targets the **Launcher**, not the running **Explorer** instance, causing several problems:

1. After accepting the login in the browser, the OS prompts the user to open the Decentraland app — even though Explorer is already running.
2. If the user **cancels** the prompt, the login in the already-open Explorer hangs (intermittent loading spinner) and never completes. There is no recovery path or feedback to the user.
3. If the user **accepts**, a new Launcher instance opens and may start downloading/updating (e.g., when the local build isn't the latest release), which is confusing.

### Solution

- When a `decentraland://` deeplink is received while an Explorer instance is already running, the launcher forwards the deeplink through the bridge file and skips the update/launch flow instead of re-running it. Applies only on `bridgeOnly` mode. The startup is now faster, showing the launcher for less than 5 seconds, removing the UX friction.
- A `&bridgeOnly` param has been added that prevents the launcher to open an Explorer instance. Useful when logging through Unity Editor.
- Pre-existent bug: Consume the deeplink from the current state, in addition to the `open_url` event. When the launcher is fired the 2nd time (on login -> use different account flow), the deeplink is already there, and the event doesn't trigger.

### Test steps

1. Download the build from this pr: https://github.com/decentraland/unity-explorer/pull/9100
2. Place it at `~/Library/Application Support/DecentralandLauncherLight/latest` (mac, windows needs another path)
3. Edit the file `'~/Application Support/DecentralandLauncherLight/config.json'
4. Add the line so the launcher doesn't update and gets overridden: `"cmd-arguments": "--never-trigger-updater"`
Example:
```
{
  "analytics-user-id": ".....",
  "cmd-arguments": "--never-trigger-updater"
}
```
5. Install this launcher
6. Close it
7. Rerun it on `zone` env as: decentraland://?dclenv=zone
8. Logout from any account
9. Login with Metamask or Google keeping the Explorer opened
10. Go to the auth website, check that the url has the `deeplink` query param
11. Sign the authentication
12. Wait until you are redirected
13. Click yes to open Decentraland
14. The launcher should open quickly and immediately redirect you to the Explorer, no update checks, no download steps
15. Check that the login flow works with multiple running instances by opening it: `decentraland://?dclenv=zone&multi-instance=true`

### DIFF
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -112,7 +112,20 @@ pub fn should_use_deeplink_bridge(
     let is_local_scene = deeplink.has_true_value(ARG_LOCAL_SCENE) || args.local_scene;
     let bridge_only = deeplink.has_true_value(ARG_BRIDGE_ONLY) || args.bridge_only;
 
-    !open_new_instance && (any_is_running || bridge_only) && !is_local_scene
+    let use_bridge = !open_new_instance && (any_is_running || bridge_only) && !is_local_scene;
+
+    log::info!(
+        "[deeplink-debug] should_use_deeplink_bridge -> {use_bridge} | deeplink={:?} | \
+         open_new_instance={open_new_instance} (multi_instance_deeplink={}, arg_open_new={}) | \
+         bridge_only={bridge_only} | is_local_scene={is_local_scene} | any_is_running={any_is_running} \
+         => {}",
+        deeplink.original(),
+        deeplink.has_true_value(ARG_MULTI_INSTANCE),
+        args.open_new_client_instance,
+        if use_bridge { "BRIDGE to running instance" } else { "LAUNCH new instance" },
+    );
+
+    use_bridge
 }
 
 pub async fn should_use_deeplink_bridge_for(
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -4,7 +4,6 @@ use std::io::Write;
 use std::time::Duration;
 use serde::Serialize;
 use tokio::fs::remove_file;
-use tokio::time::error::Elapsed;
 use tokio::time::sleep;
 use tokio_util::sync::CancellationToken;
 
@@ -115,12 +114,9 @@ pub fn should_use_deeplink_bridge(
     !open_new_instance && (any_is_running || bridge_only) && !is_local_scene
 }
 
-pub async fn should_use_deeplink_bridge_for(
-    deeplink: &DeepLink,
-    any_is_running: bool,
-) -> anyhow::Result<bool> {
+pub fn should_use_deeplink_bridge_for(deeplink: &DeepLink, any_is_running: bool) -> bool {
     let args = AppEnvironment::cmd_args();
-    Ok(should_use_deeplink_bridge(deeplink, &args, any_is_running))
+    should_use_deeplink_bridge(deeplink, &args, any_is_running)
 }
 
 pub async fn execute_passthrough<T: EventChannel>(
@@ -128,7 +124,6 @@ pub async fn execute_passthrough<T: EventChannel>(
     deeplink: &DeepLink,
 ) -> StepResult {
     const OPEN_DEEPLINK_TIMEOUT: Duration = Duration::from_secs(3);
-    type OpenResult = std::result::Result<PlaceDeeplinkResult, Elapsed>;
 
     channel.send(Status::State {
         step: Step::DeeplinkOpening,
@@ -144,33 +139,40 @@ pub async fn execute_passthrough<T: EventChannel>(
 
     let token = CancellationToken::new();
 
-    match tokio::time::timeout(
-        OPEN_DEEPLINK_TIMEOUT,
-        place_deeplink_and_wait_until_consumed(deeplink.clone(), token.child_token()),
-    )
-    .await
-    {
-        OpenResult::Ok(result) => match result {
-            PlaceDeeplinkResult::Ok(outcome) => {
-                // Once the deeplink is placed and consumed, activate the packaged Explorer window,
-                // but skip it in bridge-only mode (the consumer owns focus).
-                if outcome == DeeplinkConsumeOutcome::Consumed && bring_to_front {
-                    #[cfg(target_os = "macos")]
-                    try_bring_explorer_to_front();
-                }
-                StepResult::Ok(())
-            }
-            PlaceDeeplinkResult::Err(error) => match error {
-                PlaceDeeplinkError::SerializeError | PlaceDeeplinkError::IOError => {
-                    StepResult::Err(error.into())
-                }
-                PlaceDeeplinkError::Cancelled => StepResult::Ok(()),
-            },
-        },
-        OpenResult::Err(_) => {
+    // Pin the wait future locally so it is NOT dropped when the timer fires. That keeps its
+    // `select!` loop alive, so cancelling the token drives its own cancel branch (and cleanup).
+    let mut wait = std::pin::pin!(place_deeplink_and_wait_until_consumed(
+        deeplink.clone(),
+        token.child_token(),
+    ));
+
+    let result = tokio::select! {
+        r = &mut wait => r,
+        () = sleep(OPEN_DEEPLINK_TIMEOUT) => {
+            // Future is still alive: cancelling wakes its `token.cancelled()` arm, and awaiting it
+            // to completion lets that arm remove the bridge file before we report the timeout.
             token.cancel();
-            StepResult::Err(StepError::E3001_OPEN_DEEPLINK_TIMEOUT)
+            let _ = (&mut wait).await;
+            return StepResult::Err(StepError::E3001_OPEN_DEEPLINK_TIMEOUT);
         }
+    };
+
+    match result {
+        PlaceDeeplinkResult::Ok(outcome) => {
+            // Once the deeplink is placed and consumed, activate the packaged Explorer window,
+            // but skip it in bridge-only mode (the consumer owns focus).
+            if outcome == DeeplinkConsumeOutcome::Consumed && bring_to_front {
+                #[cfg(target_os = "macos")]
+                try_bring_explorer_to_front();
+            }
+            StepResult::Ok(())
+        }
+        PlaceDeeplinkResult::Err(error) => match error {
+            PlaceDeeplinkError::SerializeError | PlaceDeeplinkError::IOError => {
+                StepResult::Err(error.into())
+            }
+            PlaceDeeplinkError::Cancelled => StepResult::Ok(()),
+        },
     }
 }
 
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -190,17 +190,14 @@ pub async fn place_deeplink_and_wait_until_consumed(
 
 #[cfg(test)]
 mod tests {
+    #![allow(clippy::unwrap_used)]
+
     use super::*;
     use crate::environment::Args;
     use crate::protocols::DeepLink;
-    use std::collections::HashMap;
 
-    fn deeplink(flags: &[(&str, &str)]) -> DeepLink {
-        let mut map: HashMap<String, String> = HashMap::new();
-        for (key, value) in flags {
-            map.insert((*key).to_owned(), (*value).to_owned());
-        }
-        DeepLink::from_args(map)
+    fn deeplink(value: &str) -> DeepLink {
+        DeepLink::from_string(value.to_string()).unwrap()
     }
 
     fn args(argv: &[&str]) -> Args {
@@ -210,7 +207,7 @@ mod tests {
     #[test]
     fn uses_bridge_when_running_and_no_special_flags() {
         assert!(should_use_deeplink_bridge(
-            &deeplink(&[]),
+            &deeplink("decentraland://"),
             &args(&["app"]),
             true
         ));
@@ -219,7 +216,7 @@ mod tests {
     #[test]
     fn no_bridge_when_no_instance_running() {
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[]),
+            &deeplink("decentraland://"),
             &args(&["app"]),
             false
         ));
@@ -228,12 +225,12 @@ mod tests {
     #[test]
     fn no_bridge_when_new_instance_requested_via_deeplink() {
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[(ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE, "true")]),
+            &deeplink(&format!("decentraland://{}=true", ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE)),
             &args(&["app"]),
             true
         ));
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[(ARG_MULTI_INSTANCE, "true")]),
+            &deeplink(&format!("decentraland://{}=true", ARG_MULTI_INSTANCE)),
             &args(&["app"]),
             true
         ));
@@ -242,7 +239,7 @@ mod tests {
     #[test]
     fn no_bridge_when_new_instance_requested_via_args() {
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[]),
+            &deeplink("decentraland://"),
             &args(&["app", "--open-deeplink-in-new-instance"]),
             true
         ));
@@ -251,12 +248,12 @@ mod tests {
     #[test]
     fn no_bridge_for_local_scene() {
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[(ARG_LOCAL_SCENE, "true")]),
+            &deeplink(&format!("decentraland://{}=true", ARG_LOCAL_SCENE)),
             &args(&["app"]),
             true
         ));
         assert!(!should_use_deeplink_bridge(
-            &deeplink(&[]),
+            &deeplink("decentraland://"),
             &args(&["app", "--local-scene"]),
             true
         ));
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -190,14 +190,12 @@ pub async fn place_deeplink_and_wait_until_consumed(
 
 #[cfg(test)]
 mod tests {
-    #![allow(clippy::unwrap_used)]
-
     use super::*;
     use crate::environment::Args;
-    use crate::protocols::DeepLink;
+    use crate::protocols::{DeepLink, DeepLinkCreateError};
 
-    fn deeplink(value: &str) -> DeepLink {
-        DeepLink::from_string(value.to_string()).unwrap()
+    fn deeplink(value: &str) -> Result<DeepLink, DeepLinkCreateError> {
+        DeepLink::from_string(value.to_string())
     }
 
     fn args(argv: &[&str]) -> Args {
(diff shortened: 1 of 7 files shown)


core/src/flow.rs
```diff
@@ -73,17 +74,23 @@ pub struct LaunchFlow {
...
+    deeplink_passthrough_step: DeeplinkPassthroughStep,
     app_launch_step: AppLaunchStep,
 
     analytics: Arc<Mutex<Analytics>>,
 }
 
 impl LaunchFlow {
     pub fn new(
-        installs_hub: Arc<Mutex<InstallsHub>>,
+        installs_hub: &Arc<Mutex<InstallsHub>>,
```
NickKhalow: Arc<T> is already a reference, that doesn't make a sense to pass it by reference itself

core/src/flow.rs
```diff
@@ -93,14 +100,28 @@ impl LaunchFlow {
                 analytics: analytics.clone(),
                 running_instances: running_instances.clone(),
             },
-            app_launch_step: AppLaunchStep {
-                installs_hub,
-                running_instances,
+            deeplink_passthrough_step: DeeplinkPassthroughStep {
+                app_launch_step: app_launch_step.clone(),
```
NickKhalow: Step must not have ownership over an other step, should pass the dependencies directly without wrapping into an inner step

core/src/flow.rs
```diff
@@ -442,16 +474,113 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
     }
 }
 
+#[derive(Clone)]
```
NickKhalow: must not be cloneable and should exist in a single instance per run

core/src/flow.rs
```diff
@@ -442,16 +474,113 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
...
+            .execute_passthrough_internal(channel, &deeplink)
+            .await?;
+        StepResultTyped::Ok(true)
+    }
+}
+
+/// Whether an incoming deeplink should be handed off to an already-running Explorer
+/// through the file bridge instead of launching a fresh client.
+///
+/// This is the exact condition that also puts the launcher into windowless pass-through
```
NickKhalow: Comments must not explain behaviour of outer scope, it breaks the composition root

> This is the exact condition that also puts the launcher into windowless pass-through

core/src/flow.rs
```diff
@@ -566,3 +660,76 @@ impl WorkflowStep<LaunchFlowState, ()> for AppLaunchStep {
   const shouldRunDevVersion = getRunDevVersion();
   const customDownloadedFilePath = getDownloadedFilePath();
 */
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::collections::HashMap;
+
+    fn deeplink(flags: &[(&str, &str)]) -> DeepLink {
```
NickKhalow: fn should be part of DeepLink struct, `DeepLink::from_slices `

core/src/flow.rs
```diff
@@ -566,3 +660,76 @@ impl WorkflowStep<LaunchFlowState, ()> for AppLaunchStep {
...
+        let mut map: HashMap<String, String> = HashMap::new();
+        for (key, value) in flags {
+            map.insert((*key).to_owned(), (*value).to_owned());
+        }
+        DeepLink::from_args(map)
+    }
+
+    fn args(argv: &[&str]) -> Args {
+        Args::parse(argv.iter().map(|s| (*s).to_owned()))
```
NickKhalow: same as above, `Args::from_slice`

core/src/protocols.rs
```diff
@@ -72,6 +72,14 @@ impl DeepLink {
     pub fn original(&self) -> &str {
         &self.original
     }
+
+    #[cfg(test)]
+    pub(crate) const fn from_args(args: HashMap<String, String>) -> Self {
+        Self {
+            original: String::new(),
```
NickKhalow: This creates an invalid DeepLink state. original is part of the invariant and should not be filled with a placeholder value

src-tauri/src/lib.rs
```diff
@@ -83,6 +83,22 @@ async fn launch(
     launch_internal(app, state, channel).await
 }
 
+/// Show and focus the launcher's main window. The window starts hidden (see
+/// `tauri.conf.json`) so that pure deeplink pass-throughs to a running Explorer stay
```
NickKhalow: Comment explains outer behaviour:

`so that pure deeplink pass-throughs to a running Explorer stay windowless; every other path reveals it here.`

lorux0:
@claude review

core/src/environment.rs
```diff
@@ -22,6 +22,7 @@ pub const ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE: &str = "open-deeplink-in-new-instan
 // Alias of ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE: either flag enables the same behavior.
 pub const ARG_MULTI_INSTANCE: &str = "multi-instance";
 pub const ARG_LOCAL_SCENE: &str = "local-scene";
+pub const ARG_BRIDGE_ONLY: &str = "bridge-only";
```
NickKhalow: Makes a sense to add a small description for the flag, it's hard to tell what it does from the first glance

core/src/flow.rs
```diff
@@ -93,14 +100,28 @@ impl LaunchFlow {
...
-                running_instances,
+            deeplink_passthrough_step: DeeplinkPassthroughStep {
+                app_launch_step: app_launch_step.clone(),
             },
+            app_launch_step,
             analytics,
         }
     }
 
+    pub async fn is_deeplink_passthrough(&self) -> bool {
```
NickKhalow: flow.rs grows large at this point. Let's extract the deeplink bridge logic into `deeplink_bridge.rs`.

core/src/flow.rs
```diff
@@ -160,6 +181,17 @@ impl LaunchFlow {
         channel: &T,
         state: Arc<Mutex<LaunchFlowState>>,
     ) -> StepResult {
+        let handled_by_passthrough = self
+            .deeplink_passthrough_step
+            .execute_if_needed(channel, state.clone(), "launch")
+            .await?;
+        // If another Explorer instance is already running, treat this as a deeplink-only
```
NickKhalow, line 167: In case of "double activation" the return must be explicitly logged, otherwise it makes the execution harder to debug

core/src/flow.rs
```diff
@@ -442,16 +474,114 @@ impl WorkflowStep<LaunchFlowState, ()> for InstallStep {
...
+    }
+}
+
+/// Whether an incoming deeplink should be handed off to an already-running Explorer
+/// through the file bridge instead of launching a fresh client.
+///
+/// This is the exact condition that also puts the launcher into windowless pass-through
+/// mode: the user clicked a `decentraland://` link while a client is running, and they
+/// did not ask for a new instance or a local scene.
+fn should_use_deeplink_bridge(deeplink: &DeepLink, args: &Args, any_is_running: bool) -> bool {
```
NickKhalow: Same, extract to the deeplink bridge file

NickKhalow (changes requested):
Good work! I left the comments below. Overall comment: the LaunchFlow architecture gets complicted because it has to handle multiple scenarios (normal flow, deeplink passthrough). Deeplink passthrough must be extracted to DeepLinkPassthroughFlow eventually. Please address it in this PR or open a ticket to tackle it down later

core/src/flow.rs
lorux0: Fixed here: https://github.com/decentraland/launcher-rust/pull/293/changes/8bf43bcea97f3b65d9a11efcf12ac520cd3c61e0

core/src/flow.rs
lorux0: Moved everything into its own step here: https://github.com/decentraland/launcher-rust/pull/293/changes/8bf43bcea97f3b65d9a11efcf12ac520cd3c61e0

core/src/flow.rs
lorux0: Fixed: https://github.com/decentraland/launcher-rust/pull/293/changes/671060b5c60c2e4977145b25d2a22bd8480c0894

core/src/flow.rs
lorux0: Fixed: https://github.com/decentraland/launcher-rust/pull/293/changes/671060b5c60c2e4977145b25d2a22bd8480c0894

core/src/flow.rs
lorux0: No need anymore, i dont create the deeplink from a map anymore, but from a string: https://github.com/decentraland/launcher-rust/pull/293/changes/3b3d4bd5b0fbd8c082f5be359270aee8ec72496a

core/src/flow.rs
lorux0: Same, create deeplink from strings here: https://github.com/decentraland/launcher-rust/pull/293/changes/3b3d4bd5b0fbd8c082f5be359270aee8ec72496a

core/src/protocols.rs
lorux0: Changed how i create the deeplink on tests, its now done through a string: https://github.com/decentraland/launcher-rust/pull/293/changes/3b3d4bd5b0fbd8c082f5be359270aee8ec72496a

src-tauri/src/lib.rs
lorux0: Fixed: https://github.com/decentraland/launcher-rust/pull/293/changes/4e74f0af7b685a88c188cfa07a396859e83e27a3

core/src/environment.rs
lorux0: Added: https://github.com/decentraland/launcher-rust/pull/293/changes/b28ddd8ae1aec0fb4ab637bc17b40087b2dd0b8c

core/src/flow.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/pull/293/changes/777fe206e772317b48ea51b6fb17222807d01c8b

core/src/flow.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/pull/293/changes/ee0e3c4c8ebee12700fab921efbaef1504c47a04

core/src/flow.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/commit/777fe206e772317b48ea51b6fb17222807d01c8b

core/src/deeplink_bridge.rs
```diff
@@ -82,6 +92,63 @@ fn try_bring_explorer_to_front() {
...
+    let is_local_scene = deeplink.has_true_value(ARG_LOCAL_SCENE) || args.local_scene;
+    let bridge_only = deeplink.has_true_value(ARG_BRIDGE_ONLY) || args.bridge_only;
+
+    !open_new_instance && (any_is_running || bridge_only) && !is_local_scene
+}
+
+pub async fn should_use_deeplink_bridge_for(
+    deeplink: &DeepLink,
+    any_is_running: bool,
+) -> anyhow::Result<bool> {
```
NickKhalow: `anyhow::Result<bool>` should not be used here. Err value is never returned

Ludmilafantaniella (approved):
Tested on Windows and Mac. ✅

- Deep-link login flow works end-to-end: browser redirects back to the running Explorer without opening a second Launcher window ✅
- Login with Metamask and Google both complete correctly ✅
- Login cancellation returns to the login selection screen and retry works ✅
- Second sign-in after cancellation still receives the deep link ✅
- Web3 operations (Donations, Gifts) still work correctly after refactored socket signature path ✅
- Multi-instance flow (`--multi-instance`) working on both platforms ✅
- Gifts working on both platforms ✅
- Found a bug during testing: a second Launcher instance was opening when switching accounts instead of reusing the existing one — fixed by Nico and verified on both platforms ✅

LGTM 🚀

Windows Evidence:

https://github.com/user-attachments/assets/c8350698-3fb0-4a8f-a617-fd78159ee85a

Mac Evidence:

https://github.com/user-attachments/assets/593755cf-2617-4a3d-90fa-b0536ff4a2db

DONATIONS EVIDENCE:

https://github.com/user-attachments/assets/81a9088a-586e-4e27-9647-e7bc0d1ba070

GIFT EVIDENCE:

https://github.com/user-attachments/assets/dda171f2-8c38-44e6-94ca-1f4b510c555f

core/src/environment.rs
```diff
@@ -22,6 +22,9 @@ pub const ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE: &str = "open-deeplink-in-new-instan
 // Alias of ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE: either flag enables the same behavior.
 pub const ARG_MULTI_INSTANCE: &str = "multi-instance";
 pub const ARG_LOCAL_SCENE: &str = "local-scene";
+/// When present, run in bridge-only mode (no client UI).
+/// Used for deeplink/file-bridge integrations and headless operations.
+pub const ARG_BRIDGE_ONLY: &str = "bridgeOnly";
```
NickKhalow, line 27: Should use kebab case `bridgeOnly`

core/src/deeplink_bridge.rs
```diff
@@ -82,6 +101,79 @@ fn try_bring_explorer_to_front() {
...
+                StepResult::Ok(())
+            }
+            PlaceDeeplinkResult::Err(error) => match error {
+                PlaceDeeplinkError::SerializeError | PlaceDeeplinkError::IOError => {
+                    StepResult::Err(error.into())
+                }
+                PlaceDeeplinkError::Cancelled => StepResult::Ok(()),
+            },
+        },
+        OpenResult::Err(_) => {
```
NickKhalow: timeout leaves deeplink-bridge.json on disk. Is it intendent?

src-tauri/src/lib.rs
```diff
@@ -93,15 +106,31 @@ async fn launch_internal(
 
     let flow_state = guard.state.clone();
 
-    if let Err(e) = update_if_needed_and_restart(&app, &guard, &status_channel).await {
-        error!("Cannot update the launcher: {}", e);
+    // When a deeplink arrives while the Explorer is already running, run windowless:
+    // don't show the launcher window and don't check for launcher updates — just pass
+    // the deeplink through to the running client.
+    let windowless = guard.flow.is_deeplink_passthrough().await;
```
NickKhalow: TOCTOU issue

NickKhalow (changes requested):
Hey Nico!

There is major TOCTOU issue in the windowless mode. lib.rs calculates the windowless mode once and then DeeplinkPassthroughStep rechecks any_is_running() later. If the Explorer exits in between, the launcher runs a full download/install/launch with the window hidden and nothing ever reveals it

core/src/environment.rs
lorux0: Can't do, already implemented like this on the auth website

src-tauri/src/lib.rs
lorux0: Removed the windowless feature here: https://github.com/decentraland/launcher-rust/pull/293/changes/b4c9f654c8665e803b323a07e211d43ba81f4e65
Its not needed anymore as the flow now have changed and the launcher does not look for updates when a running instance is found, so in terms of UX is way faster, no issues on showing the launcher briefly for less than 5 seconds

core/src/deeplink_bridge.rs
lorux0: Fixed how the flow is resolved. The cancelled token was not working as intended, as the `Timeout` already was consuming the future, making the cancellation path invalid. Reworked into a select timer task: https://github.com/decentraland/launcher-rust/pull/293/changes/1157c8e526b197911a4ae5ed540064e3609744fe

core/src/deeplink_bridge.rs
lorux0: Fixed: https://github.com/decentraland/launcher-rust/pull/293/changes/1157c8e526b197911a4ae5ed540064e3609744fe

core/src/flow.rs
```diff
@@ -1,11 +1,8 @@
 use crate::channel::EventChannel;
-use crate::deeplink_bridge::{
-    PlaceDeeplinkError, PlaceDeeplinkResult, place_deeplink_and_wait_until_consumed,
-};
-use crate::environment::{ARG_LOCAL_SCENE, ARG_MULTI_INSTANCE, ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE};
+use crate::deeplink_bridge::{execute_passthrough, should_use_deeplink_bridge_for as should_use_bridge_for_deeplink};
```
NickKhalow: What is a purpose for `should_use_deeplink_bridge_for as should_use_bridge_for_deeplink` alias? IMO I don't see benefits, It should have a single name

core/src/flow.rs
```diff
@@ -160,6 +160,20 @@ impl LaunchFlow {
         channel: &T,
         state: Arc<Mutex<LaunchFlowState>>,
     ) -> StepResult {
+        let handled_by_passthrough = self
+            .deeplink_passthrough_step
+            .execute_if_needed(channel, state.clone(), "launch")
```
NickKhalow: Should have a unique label, currently both steps use "launch" that can confuse logs at debugging

core/src/deeplink_bridge.rs
```diff
@@ -82,6 +100,82 @@ fn try_bring_explorer_to_front() {
...
+                #[cfg(target_os = "macos")]
+                try_bring_explorer_to_front();
+            }
+            StepResult::Ok(())
+        }
+        PlaceDeeplinkResult::Err(error) => match error {
+            PlaceDeeplinkError::SerializeError | PlaceDeeplinkError::IOError => {
+                StepResult::Err(error.into())
+            }
+            PlaceDeeplinkError::Cancelled => StepResult::Ok(()),
```
NickKhalow: Looks like `PlaceDeeplinkError::Cancelled` is a dead branch, because token seems to be canceled only at 154'th line and that branch has early return

src-tauri/src/lib.rs
```diff
@@ -239,6 +239,21 @@ fn setup_deeplink(a: &App, protocol: &Protocol) {
 
     #[cfg(target_os = "macos")]
     {
+        // The deep-link plugin buffers the URL the app was launched with. When the
```
NickKhalow: What is `plugin` in that context?

NickKhalow (changes requested):
General comment: since windowsless check is removed [here](https://github.com/decentraland/launcher-rust/pull/293/changes/b4c9f654c8665e803b323a07e211d43ba81f4e65) the launcher update restarts unconditionally. It doesn't seem guaranteed that the deeplink will survive the auto-update

lorux0:
> General comment: since windowsless check is removed [here](https://github.com/decentraland/launcher-rust/pull/293/changes/b4c9f654c8665e803b323a07e211d43ba81f4e65) the launcher update restarts unconditionally. It doesn't seem guaranteed that the deeplink will survive the auto-update

What do you mean? The windowless option was removed, leaving the flow as it was originally. Maybe a pre-existent bug?

NickKhalow (approved):

NickKhalow:
> What do you mean? The windowless option was removed, leaving the flow as it was originally

PR claims: `Prevent the launcher to execute the update process when a running instance is found`, please update the description to match the new implementation

PR #295, opened 2026-07-16
feat: location from installer referrer metadata

lorux0:
## Problem

New users clicking **scene or event links** don't land where they were promised:

- **Scene links** → forced download → dropped into Genesis Plaza instead of the scene.
- **Event links** → partial flow → download completes but still lands in GP, not the event.

These flows only work for **returning users**, since deep links are only handled once the Explorer is already installed. For a first-time user, the intended location is lost somewhere in the download → install → launch funnel.

## Solution

Carry the user's intended location all the way through the funnel so first-time users land where they clicked.

- The launcher now captures the destination (position/realm) from the installer's download origin on both macOS and Windows.
- That location is persisted so it survives the download, install, and first launch.
- On first launch it's handed to the Explorer as a deeplink, so the user opens directly at the intended scene or event.
- An explicit deeplink from the user always takes priority, and the saved location is used only once — it's cleared after a successful launch so it isn't replayed later.

## Test instructions

1. Uninstall Decentraland from your device, check that deeplink `decentraland://` should not work for you anymore
2. Go to any scene coord at Genesis City on the what's on website on `zone`: https://decentraland.zone/whats-on?position=120%2C-20
3. Click jumpin, that should take you to download client
4. Download the installer and run it
5. Go through the process (launcher+explorer init)
6. Check that you started at the desired location
7. Repeat the same steps but with a world: https://decentraland.zone/whats-on?world=brai
8. Check any other jumpin to a location from the website, the flow should work the same

## Related links

https://github.com/decentraland/sites/pull/682
https://github.com/decentraland/client-download-gateway/pull/96


DafGreco (approved):

Evidence of the implementation :

https://github.com/user-attachments/assets/aca85f99-6466-4330-913a-43c0a51a1a8c

https://github.com/user-attachments/assets/a429a6bd-e955-4728-9c8c-115cc00f0f0d

NickKhalow (approved):

PR #305, opened 2026-07-23
feat: build universal macOS binary for Intel + Apple Silicon

charly-bg:
### Problem
The release built only the host architecture (arm64), so Intel Macs had no
installer and could not self-update - the live `latest.json` had no
`darwin-x86_64` key.

### Changes

**`release.yml`**
- Build macOS with `--target universal-apple-darwin` (lipo'd fat binary).
- Fix dead `matrix.platform` -> `matrix.os` so both `aarch64-apple-darwin`
  and `x86_64-apple-darwin` are installed via rustup.
- Replace the empty `secrets.args` with explicit per-platform `matrix.args`
  (Build Release and Build Dry-Run).

**`upload-to-s3/action.yml`**
- Resolve the `universal-apple-darwin` bundle dir so `Decentraland.app.tar.gz`
  is actually uploaded to S3.
- Repoint `darwin-x86_64.url` (which `tauri-action` auto-emits for universal
  builds) to the S3 artifact so Intel installs self-update; `has()`-guarded.
- Merge `latest.json` with any manifest the sibling matrix job already uploaded
  before writing to S3, so concurrent per-platform uploads no longer clobber
  each other (addresses PR review P2).

### Verification before merge
Run a dry-run that dumps `./latest.json` before the jq rewrite to confirm the
published manifest carries both darwin arches and the windows key.

### Notes
- Artifact counts are unchanged (6 release assets, 5 S3 objects); only the macOS
  release-asset arch token changes `aarch64` -> `universal`. S3 names unchanged.
- Residual: a fully race-free manifest would need a single post-matrix writer
  job; not used here because dry-run creates no draft release to merge from.

### DIFF
--- a/.github/actions/upload-to-s3/action.yml
+++ b/.github/actions/upload-to-s3/action.yml
@@ -112,6 +112,21 @@ runs:
           mv ./latest.json.tmp ./latest.json
           echo "Updated URLs in latest.json:"
           cat ./latest.json
+
+          # The macOS and Windows jobs each emit a latest.json holding only their own
+          # platform keys and upload to the same S3 path, so the last writer would drop
+          # the other's platforms. Merge with any manifest the sibling job already
+          # uploaded (local keys win on overlap). Safe because target-dir is unique per
+          # commit/run, so an existing manifest there is the same version.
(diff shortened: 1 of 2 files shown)


lorux0 (approved):

nachomazzara (approved):
LGTM

PR #312, opened 2026-07-28
chore: improve funnel analytic events

lorux0:
## Improve launch funnel analytics

**Why:** We couldn't cleanly measure the user journey from launcher start to Explorer open, or tell real drop-offs apart from normal skips.

**What changed:**

- **Measure the version-check step.** The initial "fetch latest version" step now emits start/success events. It was previously untracked, so the top of the funnel was invisible.
- **Tell "already up to date" apart from "dropped off."** Returning users on the latest version skip the download/install steps. We now emit explicit *skipped* events so they're no longer miscounted as download failures.
- **Only count a launch as successful when the Explorer actually opens.** Before, a launch counted as successful the moment the process was spawned — on macOS this could report success even when the Explorer never came up. It now reports an error in that case. A failed launch also surfaces to the user instead of silently retrying (which on macOS risked opening duplicate windows).

**Result:** the funnel from launcher open → Explorer running is now measurable end to end, with skips, drop-offs, and genuine failures each distinguishable.

## Test instructions

Check the Explorer launches correctly on macOS & Windows.

### DIFF
--- a/core/src/analytics/event.rs
+++ b/core/src/analytics/event.rs
@@ -18,6 +18,9 @@ pub enum Event {
     FETCH_VERSION_SUCCESS {
         version: String,
     },
+    FETCH_VERSION_ERROR {
+        error: String,
+    },
     DOWNLOAD_VERSION {
         version: String,
     },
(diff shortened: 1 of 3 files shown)


DafGreco (approved):

Regressions for this ticket had been performed in order to verify that the normal flow is working as expected-

PR #314, opened 2026-07-30
fix: prevent sentry duplicates and improve fingerprint codes

lorux0:
## What

- Errors are grouped in Sentry by a stable error code (tag + fingerprint), so occurrences of the same failure land in one issue instead of splitting by message.
- Errors are reported once: logs can now target a destination (file / Sentry / both), so errors captured explicitly aren't duplicated by the log bridge.
- The launch path returns typed errors with their own codes instead of generic `anyhow` errors, so each launch failure mode is pivotable on its own in Sentry and gets its own user-facing message.
- `StepError` renamed to `DCLError` — it's no longer step-specific — and error construction moved onto the type itself instead of per-module helpers.

## Test instructions

Just run the launcher normally, check everything is OK

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -77,7 +77,7 @@ pub enum DCLError {
     E1006_FILE_DELETE_FAILED {
         file_path: String,
         #[source]
-        inner_error: anyhow::Error,
+        source: std::io::Error,
     },
     E1007_FILE_CREATE_FAILED {
         file_path: String,
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -258,7 +258,7 @@ impl StepError {
                 "Decentraland isn't installed correctly. Please close the launcher and open it again to reinstall it."
             }
             Self::E3010_EXPLORER_LAUNCH_FAILED { .. } => {
-                "We couldn't start Decentraland. Please close the launcher and open it again. If the problem continues, try running it as administrator."
+                "We couldn't start Decentraland. Please close the launcher and open it again."
             }
             Self::E3011_EXPLORER_PROCESS_NOT_STARTED { .. } => {
                 "Decentraland didn't start. Please make sure your antivirus isn't blocking it and try again."
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -127,6 +127,26 @@ pub enum StepError {
     E3008_EXPLORER_ALREADY_RUNNING {
         processes: Vec<String>,
     },
+    E3009_EXPLORER_NOT_INSTALLED {
+        expected_path: String,
+        version: Option<String>,
+    },
+    E3010_EXPLORER_LAUNCH_FAILED {
+        path: String,
+        #[source]
+        inner_error: anyhow::Error,
+    },
+    E3011_EXPLORER_PROCESS_NOT_STARTED {
+        path: String,
+    },
+    E3012_EXPLORER_EXITED_ON_LAUNCH {
+        exit_code: String,
+    },
+    E3013_EXPLORER_BINARY_ACCESS_FAILED {
+        path: String,
+        #[source]
+        source: std::io::Error,
+    },
 }
 
 impl StepError {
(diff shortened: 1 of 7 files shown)


core/src/errors.rs
```diff
@@ -234,6 +254,21 @@ impl StepError {
             Self::E3008_EXPLORER_ALREADY_RUNNING { .. } => {
                 "Decentraland is already running and is blocking the update. Please close it and try again."
             }
+            Self::E3009_EXPLORER_NOT_INSTALLED { .. } => {
+                "Decentraland isn't installed correctly. Please close the launcher and open it again to reinstall it."
+            }
+            Self::E3010_EXPLORER_LAUNCH_FAILED { .. } => {
```
NickKhalow, line 289: Administrator term applies only on windows, there is no Administrator in unix systems

core/src/logs.rs
```diff
@@ -3,6 +3,9 @@ use anyhow::Result;
 use log::{Metadata, Record, info};
 use sentry_log::SentryLogger;
 
+pub const LOG_TO_FILE: &str = "dest:file";
+pub const LOG_TO_SENTRY: &str = "dest:sentry";
```
NickKhalow: These 2 consts should be a single idiomatic enum

core/src/logs.rs
```diff
@@ -58,8 +61,14 @@ impl log::Log for CombinedLog {
     }
 
     fn log(&self, record: &Record) {
-        self.fern.log(record);
-        self.sentry.log(record);
+        match record.target() {
+            LOG_TO_FILE => self.fern.log(record),
+            LOG_TO_SENTRY => self.sentry.log(record),
+            _ => {
```
NickKhalow: The match should not omit unexpected cases by default, it's error prone in long-term. If the case is expected the enum should contain the dedicated case `Both`

core/src/installs.rs
```diff
@@ -3,7 +3,7 @@ use crate::analytics::event::Event;
 use crate::config;
 use crate::download_origin_metadata::startup_location_storage::StartupDeeplinkStorage;
 use crate::environment::AppEnvironment;
-use crate::errors::{StepError, StepResult};
+use crate::errors::{StepError, StepResult, StepResultTyped};
```
NickKhalow: Semantically the use-cases don't belong to the step flow and `StepResultTyped` naming becomes misleading. Let's rename the type to `DCLError` and `DCLErrorTyped` for clarity, `StepError` shifts to `DCLError` as well

core/src/installs.rs
```diff
@@ -384,6 +394,28 @@ fn as_rename_back_err(path: &Path, source: std::io::Error) -> StepError {
     }
 }
 
+fn as_stale_cleanup_err(path: &Path, source: std::io::Error) -> StepError {
```
NickKhalow: Add corresponding `from` functions to the enum would be more idiomatic. `as_stale_cleanup_err` -> `StepError::from_cleanup`, `as_launch_failed_err` -> `StepError::from_launch_failture` and e.t.c.

core/src/installs.rs
```diff
@@ -480,8 +512,11 @@ pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) ->
         .map_err(|source| StepError::E3007_VERSION_DATA_WRITE_FAILED { source })?;
 
     // Remove the downloaded file
-    fs::remove_file(file_path)?;
-    cleanup_versions(&current_version).context("Cannot clean up the old versions")?;
+    fs::remove_file(&file_path).map_err(|e| StepError::E1006_FILE_DELETE_FAILED {
+        file_path: file_path.to_string_lossy().into_owned(),
+        inner_error: e.into(),
```
NickKhalow: Since it's always a IO error the `inner_error` field doesn't need to be the loose `anyhow::Error` type

core/src/installs.rs
```diff
@@ -480,8 +512,11 @@ pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) ->
...
+    })?;
+    cleanup_versions(&current_version)?;
```
NickKhalow: CleanUp can be returned directly since the return types do match:

`pub fn install_explorer(version: &str, downloaded_file_path: Option<PathBuf>) -> StepResult {`

`fn cleanup_versions(current_version: &EntryVersion) -> StepResult {`

NickKhalow (dismissed):
Good work! Please address the comments below

core/src/errors.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/commit/378b3976600edc5b53a333e218d54165e053e7f8

core/src/logs.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/commit/378b3976600edc5b53a333e218d54165e053e7f8

core/src/logs.rs
lorux0: Target changed into enum. However we need a default option here, otherwise all of the existing logs will break. So the default option is to keep tracking errors into sentry + logging into file, like it was originally

core/src/installs.rs
lorux0: Done https://github.com/decentraland/launcher-rust/commit/059f7eab9a61b57bd5699a0baf8bba93f87f94b6

core/src/installs.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/commit/a5e21d33ea91709eac18cacdf39b4a9749a8e431

core/src/installs.rs
lorux0: Done: https://github.com/decentraland/launcher-rust/commit/c50d4c38531cce00c7894987d9424d78c39618b4

core/src/installs.rs
lorux0: Done https://github.com/decentraland/launcher-rust/commit/dd35364592faa15d6c003e3f8ed644a5e3a27924

Ludmilafantaniella (commented):
> Verified on Windows and Mac ✅
> 
> Did a version upgrade test and a fresh install on both platforms. Also ran a regression pass:
> - Opened Creator Hub preview
> - Opened local scenes
> - Scene deeplinks (with explorer closed and open)
> 
> Everything worked as expected, no issues found.

https://github.com/user-attachments/assets/79b834a3-0e04-49f5-8d37-3ac39933061c

PR #321, opened 2026-08-03
fix: read the auth token from the path so tracking params cannot shadow it

### LINKED ISSUE #320: Auth token must be read only from the URL path, and cancellations need the campaign id
# Auth token must be read only from the URL path, and cancellations need the campaign id

Two small, independent changes on the launcher side, both found while auditing the download funnel. Neither is a regression in launcher-rust code — the first was triggered by a param the website added, the second is a missing field.

## 1. `click_id` can be mistaken for the auth token

`core/src/download_origin_metadata.rs:64` falls back to scanning for a UUID when resolving the auth token. The installer URL now carries two UUIDs:

```
https://download-gateway.decentraland.org/8448424d-5234-40c5-9922-13a2870cf7de/decentraland.dmg?anon_user_id=5b0a9de5-87e1-4321-8600-a8904f1943be&click_id=50c51fed-e0b3-4d63-9c21-4e2fb96eabb2
```

The token is the **first path segment** (present here). `click_id` is a separate query param and is not token material.

**What `click_id` is:** a website-only correlation id, minted on the download button click (`src/modules/downloadClickCorrelation.ts` in `decentraland/sites`), stored in `sessionStorage` and valid for 30 minutes. It exists so the click → `download_started` join is deterministic instead of guessing by `anonymousId`. It is **not** Google's `gclid`. By the time the launcher runs it has already expired, and neither the launcher nor the Explorer needs it.

**Why it cannot simply be removed from the URL:** the download gateway consumes it server-side (`EVENT_BODY:click_id`, live since 2026-07-13) to join a click to the delivered file. Removing it would break live telemetry, so the fix belongs on the parsing side.

**Requested contract:** when the route is `identity`, read the auth token exclusively from the first path segment. Query params are never token material, regardless of their shape.

**Measurement supporting this.** `click_id` first appeared in gateway URLs on 2026-07-13/14 (share of gateway `requested` rows carrying it jumped from 4% to 73.8%). Unique anons per day in `SEGMENT.EXPLORER_V2`:

| Window | `AUTOLOGIN_FAILURE` / day | `LOGGED_IN_CACHED` / day |
|---|---|---|
| Jul 5-13 | 1-5 (mean ≈ 3.4) | 150-226 |
| Jul 16-31 | 5-15 (mean ≈ 7.5) | 145-223 |

Auto-login failures roughly doubled from that date and stayed there, while cached logins are flat. The absolute impact is small (~3-4 users/day, since only the minority of downloads on the `identity` route carry a token at all), so this is a correctness fix, not a funnel bottleneck. Nico reproduced it locally: the token is present in the URL but auto-login does not happen.

**Verification after the fix:** `AUTOLOGIN_FAILURE` per day should return to the ~3-4 baseline.

## 2. `DOWNLOAD_VERSION_CANCELLED` is the only launcher event without `campaign_anon_user_id`

The event fires and has **2,634 rows** in `SEGMENT.DCL_LAUNCHER` (so it is not a dead enum variant), but it is the only table in that schema missing `CAMPAIGN_ANON_USER_ID`:

| Event | has `CAMPAIGN_ANON_USER_ID` |
|---|---|
| `LAUNCHER_OPEN`, `DOWNLOAD_VERSION`, `DOWNLOAD_VERSION_ERROR`, `DOWNLOAD_VERSION_SUCCESS`, `INSTALL_VERSION_START/ERROR/SUCCESS`, `LAUNCH_CLIENT_START/ERROR/SUCCESS` | yes |
| **`DOWNLOAD_VERSION_CANCELLED`** | **no** |

Consequence: a user who abandons the client download cannot be connected to the web funnel, so that drop cause is invisible in any download → in-world analysis. Copying the field from `DOWNLOAD_VERSION_ERROR`'s emission site should be enough.

For context on why this matters little today but should still be fixed: for our web-attributed downloaders (Jul 17 - Aug 1) the launcher-internal steps convert at 98-99% — `download_version` 385 → success 379, install 376, launch client 381 — with **0** cancellations observed in the attributable cohort. Closing this gap is what would let us keep asserting that with confidence rather than by absence of data.

braianj:
Closes #320 (first half).

`DownloadOriginData::from_url` scanned query params **before** the path and accepted *any* param whose value looked like a UUID as the auth token. The website now also appends `click_id`, an analytics correlation id that happens to be a UUID, so it won the race and was stored as the auth token — auto-login then failed against a value that was never a token.

Reproduced as a unit test against the current code, using the exact URL Nico extracted from an installer:

```
left:  Some("50c51fed-e0b3-4d63-9c21-4e2fb96eabb2")   <- click_id
right: Some("8448424d-5234-40c5-9922-13a2870cf7de")   <- the real token, in the path
```

### The fix

- **Path first.** The download gateway's identity route carries the token as the first path segment, so no query param can shadow it.
- **Query params match by name, not by shape.** A new `AUTH_TOKEN_QUERY_KEYS` allowlist (currently just `token`) keeps the dry-run artifact URLs working — `…/Decentraland_installer.exe?token=<uuid>` — while ignoring everything else. This is an allowlist rather than a `click_id` denylist so the next tracking param the website adds cannot re-break auto-login.

**Correction to the issue:** #320 asked for "query params are never token material". That was wrong — `?token=` is a legitimate carrier for dry-run builds and there is an existing test for it. The allowlist keeps that case intact.

`click_id` cannot be dropped from the URL on the website side: the download gateway consumes it server-side (`EVENT_BODY:click_id`, live since 2026-07-13) to join a click to the delivered file, so removing it would break live telemetry. Hence the fix belongs here.

### Measured impact

Unique anons per day in `SEGMENT.EXPLORER_V2`, around the date `click_id` reached the gateway URLs (its share of gateway requests jumped from 4% to 73.8% on Jul 13/14):

| Window | `AUTOLOGIN_FAILURE` / day | `LOGGED_IN_CACHED` / day |
|---|---|---|
| Jul 5-13 | 1-5 (mean ≈ 3.4) | 150-226 |
| Jul 16-31 | 5-15 (mean ≈ 7.5) | 145-223 |

Failures roughly doubled from that date and stayed there while cached logins remained flat. Small in absolute terms (~3-4 users/day, since only the minority of downloads on the `identity` route carry a token at all), so this is a correctness fix rather than a funnel bottleneck. Expected after deploy: back to the ~3-4/day baseline.

### Verification

- 3 new tests: the click-id-vs-path case, the anonymous route where there is no token to find, and an unknown UUID-shaped param being ignored
- Both pre-existing token cases (path segment, `?token=`) still pass
- `cargo test` in `core`: 44 passed, 0 failed
- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- Formatting: the only `rustfmt` diff in the touched file is a pre-existing import ordering at line 4 (main has 30 such diffs across the crate), left alone to keep this diff scoped

The second half of #320 — adding `campaign_anon_user_id` to `DOWNLOAD_VERSION_CANCELLED`, the only event in the `DCL_LAUNCHER` schema missing it — is not in this PR.

### DIFF
--- a/core/src/download_origin_metadata.rs
+++ b/core/src/download_origin_metadata.rs
@@ -4,9 +4,9 @@ pub mod campaign_anon_user_id_storage;
 pub mod campaign_attribution_marker;
 pub mod startup_location_storage;
 
+use anyhow::Result;
 #[cfg(target_os = "macos")]
 use anyhow::anyhow;
-use anyhow::Result;
 #[cfg(target_os = "macos")]
 use std::path::{Path, PathBuf};
 
@@ -22,6 +22,11 @@ use crate::protocols::Protocol;
 use anon_user_id::AnonUserId;
 #[cfg(target_os = "macos")]
 use campaign_anon_user_id_storage::CampaignAnonUserIdStorage;
+
+/// Query params that legitimately carry the auth token. Everything else is
+/// ignored even when its value looks like a UUID — the download URL also carries
+/// unrelated ids such as the website's `click_id`.
+const AUTH_TOKEN_QUERY_KEYS: &[&str] = &["token"];
 
 /// Data extracted from a download URL — auth token, campaign anonymous user ID,
 /// and optional startup deeplink position/realm.
@@ -39,8 +44,20 @@ pub struct DownloadOriginData {
 impl DownloadOriginData {
     /// Extract both auth token and `anon_user_id` from a single URL.
     ///
-    /// The auth token is matched by UUID regex on any query param value (except
-    /// `anon_user_id`) or path segment. The `anon_user_id` is matched by key name.
+    /// The auth token is taken from the path segment first (the download
+    /// gateway's identity route, `/{token}/decentraland.dmg`), then from the
+    /// `token` query param (used by dry-run artifact URLs). The `anon_user_id`
+    /// is matched by key name.
+    ///
+    /// Both lookups are by position or by explicit key: the previous version
+    /// accepted *any* query param whose value looked like a UUID, and checked
+    /// the query before the path. The website now also sends `click_id`, a
+    /// website-only analytics correlation id that happens to be a UUID, so it
+    /// was picked up as the auth token and shadowed the real one in the path —
+    /// auto-login then failed with a token that was never a token. Matching by
+    /// name instead of by shape means the next tracking param cannot re-break
+    /// this; anything genuinely carrying a token must be added to
+    /// `AUTH_TOKEN_QUERY_KEYS`.
     pub fn from_url(url_str: &str) -> Result<Self> {
         let url = url::Url::parse(url_str)?;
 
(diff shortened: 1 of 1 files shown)


braianj:
Parking this for now, with the resolution for context.

The bug is real and was reproduced end to end with the production launcher 1.21.2: a UUID shaped `click_id` in `kMDItemWhereFroms` ended up in both token files instead of the `identityId` from the path, and the auth API returned 404 for it.

What shipped instead: decentraland/sites#714 stopped sending `click_id` in the gateway download URL, live since sites 0.51.0 and verified in the production bundle. Auto login works today without this PR.

Still worth merging eventually, because the parser accepts any query param that looks like a UUID and survives only by luck (`anon_user_id` is excluded by name, `referrer` is a `0x` address). The next UUID shaped param on that URL breaks it again, and the launcher also parses the page URL in the xattr, which forwards `utm_*` params verbatim.

No urgency: auto auth runs on the launcher inside the DMG, before any auto update, so this only reaches users once a build carrying it becomes the gateway's base binary.

Keeping it open rather than closing it because of a separate finding: Brave never writes the download URL to the xattr, so on macOS the token never arrives at all. The likely fix is the channel the anonymous id already uses on Windows, the `identityId` in the `Content-Disposition` filename plus a filename fallback in the launcher, and that lands in this same module.

PR #325, opened 2026-08-03
fix: VCRUNTIME140.dll not found on Windows

lorux0:
## Problem

On a clean Windows install, the installer fails with:

> The code execution cannot proceed because VCRUNTIME140.dll was not found.

That DLL comes from Microsoft's Visual C++ Redistributable. Windows doesn't include it out of the box, so machines that never installed it (directly, or through some other app) can't run our binaries. Most computers happen to have it, which is why this only shows up on a fresh machine or a clean VM.

The installer already had a step meant to install the redistributable, but its check never triggers on a machine that has never had it, so the download is skipped.

Screenshot 2026-08-04 112640

## Fix

Build the Windows binaries with the C runtime bundled inside them instead of loading it from the system. No external DLL, nothing to install, nothing to download during setup.

## Testing

1. Install on a clean Windows 11 VM with no Visual C++ Redistributable — no DLL error
2. Before finishing the installation and running the explorer please make sure you have the .config file with `"cmd-arguments": "--never-trigger-updater"`
- Run the launcher after installation
- Everything should go normally until explorer opens

## Note

Binaries grow by a few hundred KB, and the first CI run recompiles everything from scratch.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
+
+[target.aarch64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -280,28 +280,6 @@ jobs:
           args: ${{ matrix.args }}
           projectPath: ${{ env.PROJECT_PATH }}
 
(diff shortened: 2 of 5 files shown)


eordano (approved):

charly-bg (approved):

lorux0:
Confirmed working on VM Windows Home plain installation

https://github.com/user-attachments/assets/f72e0a98-4080-4584-ac1a-8eeb71dffbd9

PR #327, opened 2026-08-05
fix: vcruntime dynamic installation

lorux0:
## Problem

The installer has a step that installs Microsoft's Visual C++ Redistributable when the machine doesn't have it. It never runs.

The check looks for a registry value and installs when that value says "not installed". But on a machine that never had the redistributable, the value doesn't exist at all — which is not the same as saying "not installed", so the step is skipped. It only ever triggers in a case that basically never happens in practice.

Result: the machines that need the redistributable are exactly the ones that don't get it.

The launcher itself no longer cares (#325 bundles the runtime into our binaries), but the Explorer is a Unity build that does need it.

## Fix

- Install unless the registry explicitly says the redistributable is present.
- Check both registry views — the installer is 32-bit, so it was reading the wrong place on 64-bit Windows.
- Stop treating a failed download as success: the redistributable installer is only run if the download actually worked, and failures are written to the install log.
- Fix the quoting in the download command, which was malformed and would have failed on user folders containing spaces.

## Testing

- [ ] Install on a clean Windows 11 VM with no Visual C++ Redistributable
- [ ] Install log shows `Installing the Visual C++ redistributable` with no failure lines after it
- [ ] Explorer downloads and starts
- [ ] Installing on a machine that already has the redistributable skips the step

### DIFF
--- a/src-tauri/windows/hooks.nsh
+++ b/src-tauri/windows/hooks.nsh
@@ -1,12 +1,30 @@
 !macro NSIS_HOOK_POSTINSTALL
-  ReadRegDWord $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
-  ${If} $0 == 0
-    nsExec::ExecToLog 'powershell -WindowStyle Hidden -Command "& { \
-      [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; \
-      Invoke-WebRequest -Uri \"https://aka.ms/vc14/vc_redist.x64.exe\" -OutFile \"$TEMP\vc_redist.x64.exe\"; \
-    }"'
-    nsExec::ExecToLog '"$TEMP\vc_redist.x64.exe" /quiet /norestart'
-    Delete "$TEMP\vc_redist.x64.exe"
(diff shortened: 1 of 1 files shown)


lorux0:
Confirmed working on VM clean Windows 11 Home

https://github.com/user-attachments/assets/5d06f5a5-58bf-4b18-808a-156d2f8ab793

balducciv:
✅ PR reviewed and approved by QA (smoke test) on Windows 11 following the PR test instructions.
Smoke test performed to verify the installer completes and Explorer launches as expected. ✅

**Build:** `Decentraland-1.21.4-installer` (PR #327)
**OS:** Windows 11

**Test results:**

- [x] Installer ran to completion with no errors or failure dialogs
- [x] Explorer downloaded and launched successfully after install
- [x] Machine already had Microsoft Visual C++ 2015-2022 Redistributable (x64 and x86) installed prior to testing — confirms the "already installed" skip path runs cleanly with no re-install attempt or hang

**Note:** This test did not cover the core fix path (clean machine with no VC++ Redistributable), as dev confirmed a smoke test on an existing machine is sufficient for this round. No install log was available/needed for this scope — verification is based on installer behavior and post-install Explorer launch.

**Unrelated errors noted (do not affect verdict):**

None

**Verdict: PASS ✅**

balducciv (approved):

PR #329, opened 2026-08-06
fix: verify critical native-plugin files before launch

### LINKED ISSUE #8129: System.TypeLoadException: Cannot load ClearScript V8 library. Load failure information for ClearScriptV8.win-x64.dll:
Sentry Issue: [UNITY-EXPLORER-NVA](https://decentraland.sentry.io/issues/7394000040/)

```
System.TypeLoadException: Cannot load ClearScript V8 library. Load failure information for ClearScriptV8.win-x64.dll:
C:\Users\AJ_Ro\AppData\Local\DecentralandLauncherLight\latest\runtimes\win-x64\native\ClearScriptV8.win-x64.dll: mono-io-layer-error (126)
C:\Users\AJ_Ro\AppData\Local\DecentralandLauncherLight\latest\ClearScriptV8.win-x64.dll: mono-io-layer-error (126)
  File "C:/workspace/workspace/p/Explorer/Assets/DCL/Infrastructure/SceneRuntime/V8EngineFactory.cs", line 15, in Create
  File "C:/workspace/workspace/p/Explorer/Assets/DCL/Infrastructure/SceneRuntime/SceneRuntimeImpl.cs", line 52, in .ctor
  File "C:/workspace/workspace/p/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Factory/SceneRuntimeFactory.cs", line 96, in CreateBySourceCodeAsync
  File "C:/workspace/workspace/p/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Factory/SceneRuntimeFactory.cs", line 111, in CreateByPathAsync
  File "C:/workspace/workspace/p/Explorer/Assets/DCL/Infrastructure/SceneRuntime/Factory/SceneRuntimeFactory.cs", line 96, in CreateBySourceCodeAsync
...
(107 additional frame(s) were not displayed)
```
This issue was automatically created by Sentry via [Explorer Alpha Github Issue Creation](https://sentry.io/organizations/decentraland/monitors/alerts/3130340/)

eordano:
TypeLoadException: Cannot load ClearScript V8 library (ClearScriptV8.win-x64.dll), and its duplicate DllNotFoundException signature (unity-explorer#9305), both trace to per-machine install-integrity damage, not a code regression: the 29MB ClearScriptV8.win-x64.dll ships unsigned (verified: statically-linked CRT rules out a missing VC++ redistributable; PE Security Directory empty) and is a classic antivirus-quarantine target, and the same DllNotFoundException pattern recurs on other unsigned native plugins (ktx_unity, AVProVideoWinRT, enet, segment-server, sign-server) on other machines. installs.rs::is_explorer_installed only checks that the launcher .exe exists, so a quarantined or partially-installed plugin dll passes every launcher check forever and the client re-crashes on every scene load.

Add a per-platform EXPLORER_CRITICAL_FILES list (existence + non-zero size) checked both by is_explorer_installed (pre-launch) and by install_explorer (post-extraction, before version.json is written). A missing/empty critical file forces a reinstall and reports E3014_INSTALL_INTEGRITY_CHECK_FAILED with the antivirus-quarantine explanation surfaced to the user.

Gate the check with a small per-version attempt counter (integrity-check-attempts.json, budget 3, matching flow.rs's existing SILENT_ATTEMPTS_COUNT) so a hard-coded critical-file list can't wedge every future install if a plugin is legitimately renamed or removed on some later release: the first same-session failures still hard-fail and force a reinstall exactly as before (preserving the accurate antivirus message across the existing bounded retry loop), but once a version's budget is exhausted the check degrades to advisory logging and the install/launch proceeds anyway. A version bump gets a fresh budget; a version that becomes genuinely healthy again short-circuits before ever consulting the counter.

Add unit tests for the pure verification helper, the budget-gate function, the per-version marker round-trip, and the install/launch-gate integration (20 cases); validated in an isolated scratch crate pinned to core/Cargo.lock's locked versions since the sandbox cannot compile the real crate (its sentry dependency pulls in an openssl transport unavailable here) -- caught and fixed a real clippy::arithmetic_side_effects violation in the process.

Fixes decentraland/unity-explorer#8129.
Related: decentraland/unity-explorer#9305 (duplicate DllNotFoundException signature of the same install-integrity defect).

### DIFF
--- a/core/src/errors.rs
+++ b/core/src/errors.rs
@@ -147,6 +147,9 @@ pub enum DCLError {
         #[source]
         source: std::io::Error,
     },
+    E3014_INSTALL_INTEGRITY_CHECK_FAILED {
+        missing_file: String,
+    },
 }
 
 impl DCLError {
@@ -297,6 +300,9 @@ impl DCLError {
             }
             Self::E3013_EXPLORER_BINARY_ACCESS_FAILED { .. } => {
                 "We couldn't access the Decentraland files. If Decentraland is open, please close it and try again."
+            }
+            Self::E3014_INSTALL_INTEGRITY_CHECK_FAILED { .. } => {
+                "Some Decentraland files are missing or damaged. Your antivirus may have removed them. Please add Decentraland as an exception in your antivirus and try again to reinstall."
             }
         }
     }
--- a/core/src/installs.rs
+++ b/core/src/installs.rs
@@ -47,6 +47,21 @@ const EXPLORER_MAC_APP_PATH: &str = concat!("Decentraland", ".app");
 #[cfg(target_os = "windows")]
 const EXPLORER_WIN_BIN_PATH: &str = "Decentraland.exe";
 
+#[cfg(target_os = "macos")]
+const EXPLORER_CRITICAL_FILES: &[&str] = &["Decentraland.app/Contents/MacOS/Explorer"];
+
+#[cfg(target_os = "windows")]
+const EXPLORER_CRITICAL_FILES: &[&str] = &[
+    EXPLORER_WIN_BIN_PATH,
+    "UnityPlayer.dll",
+    "Decentraland_Data/Plugins/x86_64/ClearScriptV8.win-x64.dll",
+    "Decentraland_Data/Plugins/x86_64/ktx_unity.dll",
+    "Decentraland_Data/Plugins/x86_64/AVProVideoWinRT.dll",
+    "Decentraland_Data/Plugins/x86_64/enet.dll",
+    "Decentraland_Data/Plugins/x86_64/segment-server.dll",
+    "Decentraland_Data/Plugins/x86_64/sign-server.dll",
+];
+
 pub fn log_file_path() -> Result<PathBuf> {
     let mut path = PathBuf::new();
     if let Some(dir) = dirs::home_dir() {
@@ -172,7 +187,7 @@ fn get_latest_version(version_data: &Map<String, Value>) -> DCLErrorTyped<&str> 
         .ok_or(DCLError::E3003_CANT_GET_VERSION)
 }
 
-pub(crate) fn get_explorer_launch_path(version: Option<&str>) -> DCLErrorTyped<PathBuf> {
+fn get_explorer_base_path(version: Option<&str>) -> DCLErrorTyped<PathBuf> {
     let base_path = match version {
         None => explorer_latest_version_path(),
         Some("dev") => explorer_dev_version_path(),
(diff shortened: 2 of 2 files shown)


core/src/errors.rs
```diff
@@ -147,6 +147,9 @@ pub enum DCLError {
         #[source]
         source: std::io::Error,
     },
+    E3014_INSTALL_INTEGRITY_CHECK_FAILED {
+        missing_file: String,
+    },
```
braianj, line 152: `missing_file` never reaches Sentry. `Display for DCLError` (errors.rs:305-309) writes only `user_message()`, and `code()` is derived from the variant name via strum, so this field is captured and cloned twice but only ever read by the local `log::warn!` calls, which build their own string from `missing.to_string_lossy()` anyway. The practical effect is that the E3014 event in Sentry cannot tell you whether ClearScriptV8, ktx_unity or enet was the missing file, which is the diagnostic value a dedicated error code is supposed to buy.

Either drop the field, or surface it as a tag or extra in `report_attempt_error`. If you surface it, send the relative critical-file path rather than the absolute one that `base_path.join(...)` produces: the relative path is what identifies the failure, and the absolute one embeds the user's home directory.

core/src/installs.rs
```diff
@@ -371,12 +395,123 @@ fn is_app_updated(version: &str) -> bool {
...
+        // `install_explorer`) and the file is still missing -- most likely a
+        // legitimately-changed future layout rather than recoverable
+        // per-machine corruption. Don't loop: fall through and let the flow
+        // launch what is installed, with a warning instead of a block.
+        log::warn!(
+            "{code}: critical file still missing after exhausting {INTEGRITY_CHECK_ATTEMPT_BUDGET} \
+             reinstall attempts for {version_key}; treating the install as usable so the flow \
+             launches instead of reinstalling again: {missing_file}"
+        );
+        return true;
```
braianj, line 510: The attempt counter is a one-way ratchet: nothing ever resets it. Concrete sequence:

- Antivirus quarantines a plugin, three failed installs exhaust the budget for version X.
- User adds the antivirus exception, reinstalls successfully, plays for weeks on X.
- Antivirus updates its definitions and re-quarantines the same file on the same version X.
- This branch reads attempts=3, the gate returns `None`, and it reports a broken install as usable forever, without ever attempting the reinstall that previously fixed it.

Clearing the marker whenever `find_missing_critical_file` comes back clean would close this, and would also make the budget mean "consecutive failures" rather than "failures ever"

core/src/installs.rs
```diff
@@ -371,12 +395,123 @@ fn is_app_updated(version: &str) -> bool {
...
     }
+
+    log::warn!("{code}: treating install as broken, missing or empty critical file: {missing_file}");
```
braianj, line 513: This is the path the reported scenario (quarantine after a successful install) actually takes, and it reports nothing anywhere: only `install_explorer` returns E3014 and writes the counter. Here every launch does check-fails, full re-download, install passes, launch, forever, with just a local warn. Invisible in Sentry, and a launcher that redownloads the client on every start. This path should at least report to Sentry, ideally consume budget so the loop is bounded.

core/src/installs.rs
```diff
@@ -371,12 +395,123 @@ fn is_app_updated(version: &str) -> bool {
...
+    let Some(missing) = find_missing_critical_file(&base_path) else {
+        return true;
+    };
+
+    let missing_file = missing.to_string_lossy().into_owned();
+    let code = DCLError::E3014_INSTALL_INTEGRITY_CHECK_FAILED {
+        missing_file: missing_file.clone(),
+    }
+    .code();
+    let version_key = version.unwrap_or("latest");
```
braianj, line 497: `is_explorer_installed(None)` reads the counter under `"latest"` while `install_explorer` writes it under the concrete version (line 584), so the two gates can disagree about the same install.

core/src/installs.rs
```diff
@@ -371,12 +395,123 @@ fn is_app_updated(version: &str) -> bool {
...
+
+    if integrity_check_gate(read_integrity_check_attempts(version_key), INTEGRITY_CHECK_ATTEMPT_BUDGET).is_none() {
```
braianj, line 499: Over rustfmt's max_width. CI has no `cargo fmt --check` job to catch it (release.yml only runs clippy), so it needs a manual `cargo fmt`.

braianj (commented):
Good work, the diagnosis is solid and the gap is real. I left some comments inline

PR #330, opened 2026-08-06
fix: honor the client's signin deferral window before failing a deeplink

eordano:
The launcher's deeplink "bridge" hands a deeplink to an already-running Explorer by writing deeplink-bridge.json and polling for the client to delete it, giving up after 3s and retrying the whole step 3 times -- one Sentry AttemptError per attempt (LAUNCHER-RUST-3Z8 / -HN8, and the live successor -HWC: 479 events / 78 users over 7 days). The 3s x 3 budget assumes a healthy, fully-booted, actively-polling Explorer, but since unity-explorer#9100 (deep link login, dev 2026-07-17) the client intentionally leaves a signin deeplink's bridge file on disk for up to 300s while an awaiting login claims it (DeepLinkSentinel's DEFERRED_SIGNIN_LIFETIME) -- a contract mismatch, not a protocol registration failure. Boot-window and hung-Explorer cases are also undersized by the 3s wait.

- core/src/environment.rs: add ARG_SIGNIN, mirroring the client's AppArgsFlags.SIGNIN discriminator.
- core/src/protocols.rs: add DeepLink::has_non_empty_value(key), matching the client's "query value is non-null/non-empty" signin check against the launcher's own unstripped parsed_args map (the live auth-site shape decentraland://open?signin=<id> parses to key "open?signin", which a bare args.get("signin") would miss).
- core/src/deeplink_bridge.rs: raise OPEN_DEEPLINK_TIMEOUT 3s -> 15s (single attempt, covering the boot window); on timeout, if the deeplink is a signin link, log + breadcrumb and return Ok instead of E3001, leaving the bridge file in place for the client's own 300s window instead of deleting it. Every other deeplink kind keeps exact HEAD semantics (cancel, await, remove file, E3001).
- core/src/flow.rs: mark E3001_OPEN_DEEPLINK_TIMEOUT non-retryable (one attempt instead of three x 15s), and split report_attempt_error so only the final attempt calls sentry::capture_error -- earlier attempts become breadcrumbs. Cuts AttemptError volume up to 3x for every error code, and eliminates it entirely for the signin case.

Add 23 tests across core/src/protocols.rs and core/src/flow.rs (rstest matrices) pinning the signin discriminator against 12 deeplink shapes including substring-adversarial non-matches, E3001's non-retryable status against three other error codes, and the final-attempt classification matrix that drives the breadcrumb-vs-capture split.

Fixes decentraland launcher-rust Sentry issues LAUNCHER-RUST-3Z8 / LAUNCHER-RUST-HN8.

Related: decentraland/unity-explorer#9520

### DIFF
--- a/core/src/deeplink_bridge.rs
+++ b/core/src/deeplink_bridge.rs
@@ -11,7 +11,7 @@ use crate::{
     channel::EventChannel,
     environment::{
         AppEnvironment, Args, ARG_BRIDGE_ONLY, ARG_LOCAL_SCENE, ARG_MULTI_INSTANCE,
-        ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE,
+        ARG_OPEN_DEEPLINK_IN_NEW_INSTANCE, ARG_SIGNIN,
     },
     errors::{DCLError, DCLErrorResult},
     installs::deeplink_bridge_path,
@@ -121,7 +121,7 @@ pub async fn execute_passthrough<T: EventChannel>(
     channel: &T,
     deeplink: &DeepLink,
 ) -> DCLErrorResult {
-    const OPEN_DEEPLINK_TIMEOUT: Duration = Duration::from_secs(3);
+    const OPEN_DEEPLINK_TIMEOUT: Duration = Duration::from_secs(15);
 
     channel.send(Status::State {
         step: Step::DeeplinkOpening,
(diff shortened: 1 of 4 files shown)


braianj (approved):
Verified the two subtle parts: the early `return Ok(())` leaves the bridge file in place as intended (`wait` is a locally pinned future, so dropping it skips the cancel arm that would delete it), and `has_non_empty_value` needs its two-step lookup because `parsed_args` form-urlencodes everything after `://`, putting the live `decentraland://open?signin=<id>` shape under the mangled key `open?signin`. Analytics is unaffected by the breadcrumb split.

  Two non-blocking notes:

- A hung Explorer (ANR) plus a signin deeplink is now a silent no-op, so that subcase loses its Sentry signal entirely. Worth confirming that is intended.
- For non-signin deeplinks the worst-case wait before an error goes from about 9s (3 x 3s) to a single 15s. Acceptable, but not called out in the description.

lorux0 (approved):

PR #331, opened 2026-08-06
fix: stop logging an error for data-less analytics events

eordano:
The analytics `Event` enum is serde adjacently-tagged (tag = "event", content = "data"), and serde omits the "data" key entirely for unit variants -- that is the normal, documented shape, not malformed JSON. properties_from_event() in core/src/analytics/client.rs treated a missing "data" key as an error and logged
"serialized event doesn't have data property" at error level, which sentry-log ships as a Sentry event. PR #312 added the first live unit variant, Event::FETCH_VERSION_START, emitted at the top of the fetch step on every launcher start (its is_complete() always returns false), so the error now fires once per launch: 3115 events / 476 users over 7 days on LAUNCHER-RUST-HZ0.

Extract the data lookup into a pure fn event_data(value: Value) ->
Result<Map<String, Value>, String>. A missing "data" key now returns Ok(empty map) silently instead of logging; a non-object "data" value or a non-object root still returns Err and properties_from_event logs it, same as before. The Segment-observable behavior is unchanged -- unit variants have always produced an empty properties map -- only the spurious error-level log (and the one-Sentry-event-per-launch it caused) is gone.

Add three tests to the existing core/src/analytics/client.rs test module: empty map for a unit variant, field extraction for a data-carrying variant, and Err for a non-object "data" value.

Fixes decentraland launcher-rust Sentry issue LAUNCHER-RUST-HZ0.

### DIFF
--- a/core/src/analytics/client.rs
+++ b/core/src/analytics/client.rs
@@ -163,25 +163,25 @@ fn merge_static_defaults(properties: &mut Map<String, Value>, defaults: &Map<Str
     }
 }
 
+fn event_data(value: Value) -> Result<Map<String, Value>, String> {
+    match value {
+        Value::Object(mut map) => match map.remove("data") {
+            Some(Value::Object(data)) => Ok(data),
+            Some(other) => Err(format!(
+                "serialized event data is not a json object: {other:#?}"
+            )),
+            None => Ok(Map::new()),
+        },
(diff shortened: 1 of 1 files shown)


braianj (approved):
LGTM!

lorux0 (approved):

PR #332, opened 2026-08-06
fix: terminal spawn during windows installation

lorux0:
## Problem

Users installing on Windows briefly see a terminal window pop up over the installer.

## Cause

After install, a helper runs to recover the download-origin metadata (auth token, referrer, campaign id) from the installer file. When it can't read that data directly, it falls back to calling PowerShell. The helper itself runs windowless, so Windows gives the PowerShell call its own visible console window.

## Fix

Tell Windows not to create a console window for that PowerShell call. macOS is unaffected.

## Testing

- Build and tests pass.
- Manual check on Windows: run the installer, confirm no terminal appears and the auth token is still picked up.

## Notes

Branch has a couple of throwaway commits from reproducing the issue — worth squash-merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/
 *.key
 output.log
 .DS_Store
+ 
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ dist/
 *.key
 output.log
 .DS_Store
- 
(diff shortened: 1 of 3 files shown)


eordano (approved):

lorux0:
Works on Win 11 Home

https://github.com/user-attachments/assets/7adeaae2-58ad-4b42-bae7-badfedee9b65

